PR #3, opened 2024-09-20
improve perf using pydantic

MartinBelthle:
Better usage of pydantic serialization allows us to drop the usage of json built-in library which is way slower


vargastat:
ok

PR #5, opened 2024-09-26
Feat/add local binding constraints

Sigurd-Borge:
Possible to create a binding constraint as well as add terms, tmie series are created with default zeroes if not provided

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,8 @@
 -r requirements.txt
 mypy~=1.10.0
-ruff==0.4.7
+ruff~=0.4.7
 pytest-cov~=5.0.0
 requests-mock~=1.12.1
 types-requests~=2.27.1
+tox~=4.18.1
+tox-uv~=1.11.3
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,11 +1,11 @@
-absl-py==1.4.0
-numpy==1.24.4
-protobuf==4.23.3
+absl-py~=1.4.0
+numpy~=1.26.4
+protobuf~=4.23.3
 requests~=2.31.0
-pandas ~=2.2.2
-pandas-stubs ~=2.2.2
+pandas~=2.2.2
+pandas-stubs~=2.2.2
 pytest~=7.2.1
 python-dateutil~=2.9.0
-pydantic==2.7.1
+pydantic~=2.7.1
 configparser~=5.0.2
 click~=8.1.7
--- a/src/antares/model/binding_constraint.py
+++ b/src/antares/model/binding_constraint.py
@@ -14,10 +14,11 @@ from enum import Enum
 from typing import Optional, Union, List, Any, Dict
 
 import pandas as pd
-from pydantic import BaseModel, Field, model_validator
+from pydantic import BaseModel, Field, model_validator, computed_field
 from pydantic.alias_generators import to_camel
 
 from antares.tools.contents_tool import EnumIgnoreCase, transform_name_to_id
+from antares.tools.ini_tool import check_if_none
 
 DEFAULT_GROUP = "default"
 
@@ -94,6 +95,65 @@ class BindingConstraintProperties(BaseModel, extra="forbid", populate_by_name=Tr
     group: Optional[str] = None
 
 
+class BindingConstraintPropertiesLocal(BaseModel):
+    """
+    Used to create the entries for the bindingconstraints.ini file
+
+    Args:
+        constraint_name: The constraint name
+        constraint_id: The constraint id
+        properties (BindingConstraintProperties): The BindingConstraintProperties  to set
+    """
+
+    def __init__(
+        self,
+        constraint_name: str,
+        constraint_id: str,
+        properties: Optional[BindingConstraintProperties] = None,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(**kwargs)
+        properties = properties if properties is not None else BindingConstraintProperties()
+        self._constraint_name = constraint_name
+        self._constraint_id = constraint_id
+        self._enabled = check_if_none(properties.enabled, True)
+        self._time_step = check_if_none(properties.time_step, BindingConstraintFrequency.HOURLY)
+        self._operator = check_if_none(properties.operator, BindingConstraintOperator.LESS)
+        self._comments = properties.comments
+        self._filter_year_by_year = check_if_none(properties.filter_year_by_year, "hourly")
+        self._filter_synthesis = check_if_none(properties.filter_synthesis, "hourly")
+        self._group = properties.group
+
+    @computed_field  # type: ignore[misc]
+    @property
+    def list_ini_fields(self) -> dict[str, str]:
+        ini_dict = {
+            "name": self._constraint_name,
+            "id": self._constraint_id,
+            "enabled": f"{self._enabled}".lower(),
+            "type": self._time_step.value,
+            "operator": self._operator.value,
+            "comments": self._comments,
+            "filter-year-by-year": self._filter_year_by_year,
+            "filter-synthesis": self._filter_synthesis,
+            "group": self._group,
+        }
+        return {key: value for key, value in ini_dict.items() if value is not None}
+
+    @computed_field  # type: ignore[misc]
+    @property
+    def yield_binding_constraint_properties(self) -> BindingConstraintProperties:
+        return BindingConstraintProperties(
+            enabled=self._enabled,
+            time_step=self._time_step,
+            operator=self._operator,
+            comments=self._comments,
+            filter_year_by_year=self._filter_year_by_year,
+            filter_synthesis=self._filter_synthesis,
+            group=self._group,
+        )
+
+
 class BindingConstraint:
     def __init__(  # type: ignore # TODO: Find a way to avoid circular imports
         self,
@@ -107,6 +167,9 @@ class BindingConstraint:
         self._id = transform_name_to_id(name)
         self._properties = properties or BindingConstraintProperties()
         self._terms = {term.id: term for term in terms} if terms else {}
+        self._local_properties = BindingConstraintPropertiesLocal(
+            constraint_name=self._name, constraint_id=self._id, properties=properties
+        )
 
     @property
     def name(self) -> str:
@@ -119,6 +182,14 @@ class BindingConstraint:
     @property
     def properties(self) -> BindingConstraintProperties:
         return self._properties
+
+    @properties.setter
+    def properties(self, new_properties: BindingConstraintProperties) -> None:
+        self._properties = new_properties
+
+    @property
+    def local_properties(self) -> BindingConstraintPropertiesLocal:
+        return self._local_properties
 
     def get_terms(self) -> Dict[str, ConstraintTerm]:
         return self._terms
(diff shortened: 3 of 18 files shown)


src/antares/model/binding_constraint.py
```diff
@@ -45,6 +44,16 @@ class TermOperators(BaseModel):
     weight: Optional[float] = None
     offset: Optional[int] = None
 
+    @computed_field  # type: ignore[misc]
+    @property
+    def weight_offset(self) -> str:
```
MartinBelthle, line 49: I don't understand why you've introduced that

src/antares/model/binding_constraint.py
```diff
@@ -84,14 +93,73 @@ def generate_id(cls, data: Union[Dict[str, str], LinkData, ClusterData]) -> str:
...
-    time_step: Optional[BindingConstraintFrequency] = None
-    operator: Optional[BindingConstraintOperator] = None
-    comments: Optional[str] = None
-    filter_year_by_year: Optional[str] = None
-    filter_synthesis: Optional[str] = None
-    group: Optional[str] = None
+class DefaultBindingConstraintProperties(BaseModel, extra="forbid", populate_by_name=True, alias_generator=to_camel):
+    """Default properties for binding constraints
+
+    Attributes:
```
MartinBelthle, line 99: Is it necessary to add this doc as it exactly the same thing as below ?

src/antares/model/binding_constraint.py
```diff
@@ -84,14 +93,73 @@ def generate_id(cls, data: Union[Dict[str, str], LinkData, ClusterData]) -> str:
...
+        constraint_id: The constraint id
+        properties (BindingConstraintProperties): The BindingConstraintProperties  to set
+        terms (dict[str, ConstraintTerm]]): The terms applying to the binding constraint
+    """
+
+    constraint_name: str
+    constraint_id: str
+    terms: dict[str, ConstraintTerm] = {}
+
+    @computed_field  # type: ignore[misc]
```
MartinBelthle, line 139: Not sure we should add this decorator as it is used to add the property inside the serialization and inside the yield method we exclude it.

src/antares/model/binding_constraint.py
```diff
@@ -120,6 +195,14 @@ def id(self) -> str:
     def properties(self) -> BindingConstraintProperties:
         return self._properties
 
+    @properties.setter
```
MartinBelthle, line 198: Not really familiar with this but isn't it a bit weird to have both a property and a property setter ? Shouldn't be just remove the private `_` then ?

src/antares/model/study.py
```diff
@@ -265,11 +265,9 @@ def create_binding_constraint(
         equal_term_matrix: Optional[pd.DataFrame] = None,
         greater_term_matrix: Optional[pd.DataFrame] = None,
     ) -> BindingConstraint:
-        constraint = self._binding_constraints_service.create_binding_constraint(
+        return self._binding_constraints_service.create_binding_constraint(
             name, properties, terms, less_term_matrix, equal_term_matrix, greater_term_matrix
         )
-        self._binding_constraints[constraint.id] = constraint
```
MartinBelthle, line 271: Why did you remove this line ?

src/antares/service/base_services.py
```diff
@@ -391,6 +391,11 @@ def add_constraint_terms(self, constraint: BindingConstraint, terms: List[Constr
         """
         pass
 
+    @property
+    @abstractmethod
+    def binding_constraints(self) -> dict[str, BindingConstraint]:
+        pass
```
MartinBelthle, line 397: Is it normal ?

src/antares/service/local_services/binding_constraint_local.py
```diff
@@ -20,29 +21,116 @@
...
+        # Add binding constraints
+        self._binding_constraints[constraint.id] = constraint
+        self._write_binding_constraint_ini()
+
+        # Add constraint time series
+        time_series_length = (
+            (365 * 24 + 24) if constraint.properties.time_step == BindingConstraintFrequency.HOURLY else 366
+        )
+
+        if constraint.properties.operator in (BindingConstraintOperator.LESS, BindingConstraintOperator.BOTH):
```
MartinBelthle, line 67: This if else logic is pretty long, perhaps you could place it inside a method to "hide" its complexity

src/antares/tools/contents_tool.py
```diff
@@ -118,7 +118,7 @@ def to_craft(self) -> Dict[str, Any]:
 
 # TODO maybe put sorting functions together
 def sort_ini_sections(ini_to_sort: configparser.ConfigParser) -> configparser.ConfigParser:
-    sorted_ini = configparser.ConfigParser()
+    sorted_ini = configparser.ConfigParser(interpolation=None)
```
MartinBelthle, line 121: Smart

MartinBelthle (commented):
Seems legit :), just minor comments

requirements-dev.txt
```diff
@@ -1,6 +1,8 @@
 -r requirements.txt
 mypy~=1.10.0
-ruff==0.4.7
+ruff~=0.4.7
 pytest-cov~=5.0.0
 requests-mock~=1.12.1
 types-requests~=2.27.1
+tox~=4.18.1
```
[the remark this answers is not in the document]
Sigurd-Borge: Included a bit early in the requirements (see comment in commit)

[from README.md in feat/update_project_management](https://github.com/AntaresSimulatorTeam/antares_craft/blob/feat/update_project_management/README.md):

### tox
To use [tox](https://tox.wiki/) to run unit tests in multiple python versions at the same time as linting and formatting
with ruff and typing with mypy:
1) As the dev requirements include [uv](https://docs.astral.sh/uv/) and `tox-uv` there is no need to install python 
versions, `uv` will do this for you.
2) Use `tox -p` to run the environments in parallel to save time, this will create virtual environment with the 
necessary python versions the first time you run tox.

requirements.txt
```diff
@@ -1,11 +1,11 @@
-absl-py==1.4.0
-numpy==1.24.4
-protobuf==4.23.3
+absl-py~=1.4.0
+numpy~=1.26.4
+protobuf~=4.23.3
```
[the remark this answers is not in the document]
Sigurd-Borge: Don't think we do - was included in the first commit also before github as a possible tech to use for passing messages between different projects.

src/antares/model/binding_constraint.py
Sigurd-Borge: This is for the way the weight and offset are written in the .ini

src/antares/model/binding_constraint.py
Sigurd-Borge: I included it as possibly useful information for generating docs later when possibly adding [Sphinx].(https://www.sphinx-doc.org/) or similar to the project-management branch

src/antares/model/binding_constraint.py
Sigurd-Borge: Probably, I believe I might have forgotten the case where we need to update the local properties from the new properties here and that would be the reason for the setter

src/antares/model/binding_constraint.py
Sigurd-Borge: My understanding is that a property calculated from other fields like this one should be using the `@computed_field` decorator, however the code seems like it might still work without it so maybe it's not necessary even if strictly speaking "correct"?

src/antares/service/base_services.py
Sigurd-Borge: I'm sorry I don't understand this comment, could you clarify?

src/antares/service/local_services/binding_constraint_local.py
Sigurd-Borge: Yeah good call - should be extracted, it annoyed me when doing it too

src/antares/model/study.py
Sigurd-Borge: I've moved the tracking of binding constraints from the 'Study' object to the 'binding constraints service' to correctly handle the .ini file.

requirements-dev.txt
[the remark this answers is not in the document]
Sigurd-Borge: Also possible it slipped in by mistake since I've been using tox locally for awhile, either way including it is intentional in the other branch.

requirements.txt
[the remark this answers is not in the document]
vargastat: My bad, I did some tests and forgot to remove it

Sigurd-Borge:
Think I've replied to all of them.

src/antares/model/binding_constraint.py
MartinBelthle: See the doc here : https://docs.pydantic.dev/2.0/usage/computed_fields/, IMO you should remove it

requirements-dev.txt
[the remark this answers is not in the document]
MartinBelthle: Okay why not

src/antares/service/base_services.py
MartinBelthle: I was refering to the `pass`, shouldn't you do something here. Or if it's normal update the expected response model

src/antares/model/study.py
MartinBelthle: Ok

src/antares/model/binding_constraint.py
MartinBelthle: See the doc here https://docs.pydantic.dev/2.0/usage/computed_fields/. IMO you should remove it and also remove it for `list_ini_fields` properties

src/antares/model/binding_constraint.py
MartinBelthle: Ok I see. IMO you should make it a method and not a property as it's confusing.

src/antares/model/binding_constraint.py
MartinBelthle: Ok

requirements.txt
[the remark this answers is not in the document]
MartinBelthle: Ok so you can remove it :)

src/antares/service/base_services.py
Sigurd-Borge: I'm sorry, still confused. This is the abstract base service, the `pass` is pervasive throughout not just in this method. The response model is for inheritance.

requirements.txt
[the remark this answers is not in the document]
Sigurd-Borge: Removed

src/antares/model/binding_constraint.py
Sigurd-Borge: Corrected

src/antares/service/local_services/binding_constraint_local.py
Sigurd-Borge: Extracted

src/antares/model/binding_constraint.py
MartinBelthle: Cool :)

src/antares/service/base_services.py
MartinBelthle: Ok nevermind I misunderstood. But I believe you should remove the @property as it's clearly not one.

MartinBelthle (approved):

PR #6, opened 2024-09-27
Test/cicd

berthetclement:


MartinBelthle:
Not relevant anymore as we already use windows inside the CI

PR #9, opened 2024-11-05
Feat/add study settings

Sigurd-Borge:
Study settings should now work during study creation, included creation of new custom configparser taken from AntaREST and rewritten handling of playlists as well as thematic trimming.


vargastat (approved):
Nice Job! :)

PR #12, opened 2024-11-22
feat(ci): use tox

MartinBelthle:
This PR achieves several things:
- Run the CI on ubuntu20, 22 and windows with python 3.9,10,11,12 for ubuntu and 3.9 for windows (already long)
- Fix failing tests on Windows
- Start refactoring on matrices to simplify classes and to (in a next PR) avoid having TimeSeries inside objects.
This will allow us to always have "light" objects 
- Miscellaneous changes


vargastat (approved):
Good idea not to have specific properties for model elements (solar, wind) and just count in implementation to write TimeSeries :)

PR #13, opened 2024-11-26
feat(api): add read method for thermal, renewable and st_storage

mehdiwahada:


src/antares/service/api_services/renewable_api.py
```diff
@@ -63,3 +64,23 @@ def get_renewable_matrix(self, renewable: RenewableCluster) -> pd.DataFrame:
             return get_matrix(f"{self._base_url}/studies/{self.study_id}/raw?path={path}", self._wrapper)
         except APIError as e:
             raise RenewableMatrixDownloadError(renewable.area_id, renewable.name, e.message) from e
+
+    def read_renewables(
+        self,
+        area_id: str,
+    ) -> List[RenewableCluster]:
+        json_renewables = self._wrapper.get(
+            self._base_url + "/studies/" + self.study_id + "/areas/" + area_id + "/clusters/renewable"
```
MartinBelthle, line 73: Minor change for the 3 services, you could use a f-string as it's easier to read:
`url = f"{self._base_url}/studies/{self.study_id}/areas/{area_id}/clusters/renewable"`
`json_renewables = self._wrapper.get(url).json()`

MartinBelthle (changes requested):
Just a minor change. Also, could you add an integration test inside the `test_web_client.py` to ensure everything works ? (1 test for each read you've introduced)

tests/integration/test_web_client.py
```diff
@@ -183,6 +186,31 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop):
         assert storage_fr.name == st_storage_name
         assert storage_fr.id == "cluster_test"
 
+        # testing
```
MartinBelthle, line 189: Change this message to a more explicit one explaining which tests you're performing.

src/antares/service/api_services/st_storage_api.py
```diff
@@ -73,3 +74,18 @@ def get_storage_matrix(self, storage: STStorage, ts_name: STStorageMatrixName) -
...
+
+        for storage in json_storage:
+            storage_id = storage.pop("id")
+            storage_name = storage.pop("name")
+
+            storage_properties = STStorageProperties(**storage)
+            st_storage = STStorage(self.config, storage_id, storage_name, storage_properties)
+            storages.append(st_storage)
+
+        return storages
```
MartinBelthle, line 91: I forgot this but it could be nice to sort the objects based on their ids, as you did inside the `read_areas` method to ensure test reproducibility.
**Same goes for the 3 services**

tests/integration/test_web_client.py
```diff
@@ -183,6 +186,31 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop):
...
+        thermal_service = ThermalApiService(api_config, study_id=study.service.study_id)
+        renewable_service = RenewableApiService(api_config, study_id=study.service.study_id)
+        storage_service = ShortTermStorageApiService(api_config, study_id=study.service.study_id)
```
MartinBelthle, line 192: Here you shouldn't introduce these services. It's an implementation detail that the user shouldn't know about. What the user will do is **area_fr.read_thermal_clusters()** which will call the thermal service but he doesn't have to know about this. This goes for the 3 services you introduced.

MartinBelthle (approved):

PR #18, opened 2024-11-28
Added several error messages when trying to create an element that already exists

Sigurd-Borge:
Also refactored `create_study_local` to not need a "LocalConfiguration" object but just the path to the parent directory as a string.

### DIFF
--- a/src/antares/model/study.py
+++ b/src/antares/model/study.py
@@ -23,7 +23,7 @@ import pandas as pd
 from antares.api_conf.api_conf import APIconf
 from antares.api_conf.request_wrapper import RequestWrapper
 from antares.config.local_configuration import LocalConfiguration
-from antares.exceptions.exceptions import APIError, AreaCreationError, StudyCreationError
+from antares.exceptions.exceptions import APIError, StudyCreationError
 from antares.model.area import Area, AreaProperties, AreaUi
 from antares.model.binding_constraint import BindingConstraint, BindingConstraintProperties, ConstraintTerm
 from antares.model.link import Link, LinkProperties, LinkUi
@@ -31,7 +31,6 @@ from antares.model.settings.study_settings import DefaultStudySettings, StudySet
 from antares.model.settings.time_series import correlation_defaults
 from antares.service.api_services.study_api import _returns_study_settings
 from antares.service.base_services import BaseStudyService
-from antares.service.local_services.study_local import StudyLocalService
 from antares.service.service_factory import ServiceFactory
 from antares.tools.ini_tool import IniFile, IniFileTypes
 
@@ -218,8 +217,6 @@ class Study:
     def create_area(
         self, area_name: str, *, properties: Optional[AreaProperties] = None, ui: Optional[AreaUi] = None
     ) -> Area:
-        if area_name in self._areas and isinstance(self._study_service, StudyLocalService):
-            raise AreaCreationError(area_name, f"There is already an area '{area_name}' in the study '{self.name}'")
         area = self._area_service.create_area(area_name, properties, ui)
         self._areas[area.id] = area
         return area
--- a/src/antares/model/study.py
+++ b/src/antares/model/study.py
@@ -23,7 +23,7 @@ import pandas as pd
 from antares.api_conf.api_conf import APIconf
 from antares.api_conf.request_wrapper import RequestWrapper
 from antares.config.local_configuration import LocalConfiguration
-from antares.exceptions.exceptions import APIError, StudyCreationError
+from antares.exceptions.exceptions import APIError, AreaCreationError, StudyCreationError
 from antares.model.area import Area, AreaProperties, AreaUi
 from antares.model.binding_constraint import BindingConstraint, BindingConstraintProperties, ConstraintTerm
 from antares.model.link import Link, LinkProperties, LinkUi
@@ -31,6 +31,7 @@ from antares.model.settings.study_settings import DefaultStudySettings, StudySet
 from antares.model.settings.time_series import correlation_defaults
 from antares.service.api_services.study_api import _returns_study_settings
 from antares.service.base_services import BaseStudyService
+from antares.service.local_services.study_local import StudyLocalService
 from antares.service.service_factory import ServiceFactory
 from antares.tools.ini_tool import IniFile, IniFileTypes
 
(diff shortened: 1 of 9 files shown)


src/antares/service/local_services/binding_constraint_local.py
```diff
@@ -47,6 +48,10 @@ def create_binding_constraint(
         equal_term_matrix: Optional[pd.DataFrame] = None,
         greater_term_matrix: Optional[pd.DataFrame] = None,
     ) -> BindingConstraint:
+        if name in self.binding_constraints:
```
MartinBelthle, line 51: You should check with the id as it's what you store inside self.binding_constraints

src/antares/service/local_services/link_local.py
```diff
@@ -74,15 +75,19 @@ def create_link(
...
                 properties_ini.read_file(ini_file)
         try:
             properties_ini.add_section(area_to.name)
-        except configparser.DuplicateSectionError as e:
-            raise CustomError(f"Link exists already, section already exists in properties.ini:\n\n{e.message}")
+        except DuplicateSectionError as e:
+            raise LinkCreationError(
+                area_from=area_from.name,
+                area_to=area_to.name,
+                message=f"Link exists already between '{area_from.name}' and '{area_to.name}', section already exists in properties.ini:\n\n{e.message}",
```
MartinBelthle, line 89: The first part of the msg is enough I believe: f"Link exists already between '{area_from.name}' and '{area_to.name}'"

src/antares/service/local_services/area_local.py
```diff
@@ -84,7 +84,14 @@ def create_thermal_cluster(
...
 
         list_ini = IniFile(self.config.study_path, IniFileTypes.THERMAL_LIST_INI, area_name=area_id)
-        list_ini.add_section(local_thermal_properties.list_ini_fields)
+        try:
+            list_ini.add_section(local_thermal_properties.list_ini_fields)
+        except DuplicateSectionError as error:
+            raise ThermalCreationError(
+                thermal_name,
+                area_id,
+                f"A thermal cluster called '{thermal_name}' already exists in area '{area_id}'. " + error.message,
```
MartinBelthle, line 93: Same here don't need the errror.message I think

src/antares/model/study.py
```diff
@@ -77,7 +78,7 @@ def create_study_api(
 def create_study_local(
     study_name: str,
     version: str,
-    local_config: LocalConfiguration,
+    parent_directory: str,
```
MartinBelthle, line 81: I think it's better indeed

src/antares/model/study.py
```diff
@@ -216,6 +218,8 @@ def get_binding_constraints(self) -> MappingProxyType[str, BindingConstraint]:
     def create_area(
         self, area_name: str, *, properties: Optional[AreaProperties] = None, ui: Optional[AreaUi] = None
     ) -> Area:
+        if area_name in self._areas and isinstance(self._study_service, StudyLocalService):
```
MartinBelthle, line 221: Should compare id and not name. And it should be done inside the local service method create_area, not here as you introduced the notion of local inside a model which isn't supposed to have one (plus it's a bit ugly)

tests/antares/services/local_services/test_area.py
```diff
@@ -424,7 +444,7 @@ def test_st_storage_list_ini_has_correct_default_values(
         assert actual_st_storage_list_ini.parsed_ini.sections() == expected_st_storage_list_ini.sections()
         assert actual_st_storage_list_ini.parsed_ini == expected_st_storage_list_ini
 
-    def test_st_storage_and_ini_have_custom_properties(self, local_study_with_st_storage, actual_st_storage_list_ini):
+    def test_st_storage_and_ini_have_custom_properties(self, local_study_with_renewable):
```
MartinBelthle, line 447: Really weird to use _with_renewable in a st_storage test, why did you change this ?

tests/antares/services/local_services/test_area.py
```diff
@@ -52,6 +55,18 @@ def test_can_be_created(self, local_study_w_areas):
         created_thermal = local_study_w_areas.get_areas()["fr"].create_thermal_cluster(thermal_name)
         assert isinstance(created_thermal, ThermalCluster)
 
+    def test_duplicate_name_errors(self, local_study_w_thermal):
```
MartinBelthle, line 58: Duplicated test inside test_study. This one should be removed i think

MartinBelthle (changes requested):
Few changes requested

tests/antares/services/local_services/test_area.py
Sigurd-Borge: Because in this test I test that creating a short term storage with custom properties work, this way it's clear that there is no st storage in the area yet and avoids potential duplicate st storage error.

MartinBelthle: Okay so it would be better to use _w_areas just to make it clearer

vargastat (approved):

tests/antares/services/local_services/test_area.py
Sigurd-Borge: I'm a bit confused, I don't see a dupe and this seems to be the right place to test this as thermals are part of areas. Could you point me at the duplicated test please?

tests/antares/services/local_services/test_area.py
Sigurd-Borge: Corrected

src/antares/service/local_services/binding_constraint_local.py
```diff
@@ -47,6 +49,10 @@ def create_binding_constraint(
         equal_term_matrix: Optional[pd.DataFrame] = None,
         greater_term_matrix: Optional[pd.DataFrame] = None,
     ) -> BindingConstraint:
+        if transform_name_to_id(name) in self.binding_constraints:
```
MartinBelthle, line 52: Mmmmh you assume the id equals transform_name_to_id(name) but it's an implementation detail of the class BindingConstraint. You should just put this code after creating the constraint as you can use the created id without assuming anything

tests/antares/services/local_services/test_area.py
MartinBelthle: You're right I misread my bad

MartinBelthle (changes requested):
Just one minor change and we're good to go :)

MartinBelthle (approved):

PR #20, opened 2024-11-28
Features/add hydro data

mehdiwahada:


src/antares/model/area.py
```diff
@@ -373,3 +373,9 @@ def read_thermal_clusters(
         area_id: str,
     ) -> List[ThermalCluster]:
         return self._thermal_service.read_thermal_clusters(area_id)
+
+    def read_hydro(
+        self,
+        area_id: str,
```
MartinBelthle, line 379: Inside area you shouldn't have area_id as the user will do area.read_hydro()

tests/antares/services/api_services/test_area_api.py
```diff
@@ -342,3 +342,37 @@ def test_read_areas(self):
...
+            "useLeeway": "false",
+            "leewayLow": 1,
+            "leewayUp": 1,
+            "pumpingEfficiency": 1,
+        }
+        url = f"https://antares.com/api/v1/studies/{study_id_test}/areas/{area_id}/hydro/form"
+
+        with requests_mock.Mocker() as mocker:
+            mocker.get(url, json=json_hydro)
+            area_api = AreaApiService(self.api, study_id_test)
```
MartinBelthle, line 370: Here you shouldn't have to introduce a service, you should just do area.read_hydro() as it's done inside integration test.
Also you can use self.area and self.study_id instead of introducing study_id and area_id at the beginning of the test to simplify it

MartinBelthle (changes requested):

MartinBelthle (approved):

PR #21, opened 2024-12-02
feat(local): add renewable reading + various matrix reading

killian-scalian:
- Add renewable reading
- Unitary tests
-  Add timeserie services on AreaLocalService 
  - solar
  - wind
  - reserves
  - misc-gen


src/antares/model/area.py
```diff
@@ -299,6 +299,9 @@ def get_reserves_matrix(self) -> pd.DataFrame:
 
     def get_misc_gen_matrix(self) -> pd.DataFrame:
         return self._area_service.get_misc_gen_matrix(self)
+    
+    def get_renewable_matrix(self, renewable: RenewableCluster) -> pd.DataFrame:
```
MartinBelthle, line 303: You should only give a renewable_id not a renewable object

src/antares/service/local_services/renewable_local.py
```diff
@@ -29,12 +35,40 @@ def update_renewable_properties(
         self, renewable_cluster: RenewableCluster, properties: RenewableClusterProperties
     ) -> RenewableClusterProperties:
         raise NotImplementedError
+    
+    def _read_timeseries(self, ts_file_type: TimeSeriesFileType, study_path: Path, area_id: Optional[str] = None, renewable_id: Optional[str] = None) -> pd.DataFrame:
```
MartinBelthle, line 39: You should put this method inside a utils as it's a duplicate from the one inside area_local

tests/antares/services/local_services/test_area.py
```diff
@@ -1139,12 +1141,165 @@ def _write_file(_file_path, _time_series) -> None:
...
+        for area in areas:
+            expected_time_serie = pd.DataFrame(
+                    [
+                        [-9999999980506447872, 0, 9999999980506447872],
+                        [0, area.id, 0],
+                    ]
+                , dtype="object")
+            renewable_list = area.read_renewables(area.id)
+
+            if renewable_list:
```
MartinBelthle, line 1172: Weird to add a if inside a test that will always go inside or outside of it

MartinBelthle (changes requested):
Minor changes requested

src/antares/model/area.py
killian-scalian: Ok i modified the signature, and changed accordingly the signature on renewable_api as well as the unit tests associated

src/antares/model/area.py
```diff
@@ -299,6 +299,9 @@ def get_reserves_matrix(self) -> pd.DataFrame:
 
     def get_misc_gen_matrix(self) -> pd.DataFrame:
         return self._area_service.get_misc_gen_matrix(self)
+    
+    def get_renewable_matrix(self, renewable_id: str) -> pd.DataFrame:
```
MartinBelthle, line 303: I just realized this doesn't make sense here. As an user will do `renewable.get_timeseries()` and not area.get_ts(renewable_id) it doesn't make sense. So you can remove this. 
Also rename the method `get_timeseries()` as it would be clearer for the user

MartinBelthle (changes requested):

src/antares/model/area.py
killian-scalian: Okay understood, i removed it and renamed get_renewable_matrxi from renewableCluster object

src/antares/tools/matrix_tool.py
```diff
@@ -36,3 +40,17 @@ def df_save(df: pd.DataFrame, path: Path) -> None:
 
 def df_read(path: Path) -> pd.DataFrame:
     return pd.read_csv(path, sep="\t", header=None)
+
+
+def read_timeseries(ts_file_type: TimeSeriesFileType, study_path: Path, area_id: Optional[str] = None, constraint_id: Optional[str] = None, renewable_id: Optional[str] = None,) -> pd.DataFrame:
```
MartinBelthle, line 45: As st_storage, thermal and renewable are all clusters, it could be nice to rename this parameter `cluster_id` otherwise you'll have to add other arguments here. This goes with changing `RENEWABLE_DATA_SERIES` to put `cluster_id` inside it. 

After this it will be good for me

MartinBelthle (changes requested):

src/antares/tools/matrix_tool.py
killian-scalian: No problem, i changed mentions of renewable_id for cluster_id, for more clarity

MartinBelthle (approved):

PR #24, opened 2024-12-03
ci(lint): forbid wrong typing inside the CI

Sigurd-Borge:
Current tox.ini and ci.yml doesn't actually fail if files are incorrectly written (it tries to fix with `ruff check --fix`) or formatted according to the rules in ruff.toml, this corrects that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Performs Ubuntu tests
         if: matrix.os != 'windows-latest'
         run: |
-          tox -e lint-github
+          tox -e lint-ci
           tox -p
 
       - name: Performs Windows tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,9 @@ jobs:
 
       - name: Performs Ubuntu tests
         if: matrix.os != 'windows-latest'
-        run: tox -p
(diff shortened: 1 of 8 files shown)


tox.ini
```diff
@@ -27,3 +26,10 @@ commands =
     ruff format src/ tests/ {posargs}
     mypy {posargs}
 
+[testenv:lint-github]
+description = Linting and formatting with ruff, typing with mypy
```
MartinBelthle, line 30: You should add the licenser checker with --action=check-strict

.github/workflows/ci.yml
```diff
@@ -29,7 +29,9 @@ jobs:
 
       - name: Performs Ubuntu tests
         if: matrix.os != 'windows-latest'
-        run: tox -p
+        run: |
+          tox -e lint-github
```
MartinBelthle, line 33: Change this to lint-ci ?

MartinBelthle (changes requested):

tox.ini
Sigurd-Borge: Like minds, already done ;)

.github/workflows/ci.yml
Sigurd-Borge: If you want :)

MartinBelthle (approved):

PR #25, opened 2024-12-04
BindingConstraint service refactored

vargastat:
binding_constraints removed from serviced and moved to study level

### DIFF
--- a/src/antares/model/binding_constraint.py
+++ b/src/antares/model/binding_constraint.py
@@ -91,9 +91,6 @@ class ConstraintTerm(TermOperators):
             return "%".join(sorted((data.area1.lower(), data.area2.lower())))
         return ".".join((data.area.lower(), data.cluster.lower()))
 
-    def serialize_term_data(self) -> Optional[str]:
-        return self.weight_offset()
-
 
 class DefaultBindingConstraintProperties(BaseModel, extra="forbid", populate_by_name=True, alias_generator=to_camel):
     """Default properties for binding constraints
--- a/src/antares/model/binding_constraint.py
+++ b/src/antares/model/binding_constraint.py
@@ -91,6 +91,19 @@ class ConstraintTerm(TermOperators):
             return "%".join(sorted((data.area1.lower(), data.area2.lower())))
         return ".".join((data.area.lower(), data.cluster.lower()))
 
+    def serialize_term_data(self) -> Optional[str]:
+        """
+        Serializes the term data to be correctly written in INI.
+        """
+        if isinstance(self.data, LinkData):
+            if self.offset is not None:
+                return f"0.000000%{self.offset}"
+            if self.weight is not None:
+                return f"{self.weight}"
+            return "0"
+
+        return None
+
 
 class DefaultBindingConstraintProperties(BaseModel, extra="forbid", populate_by_name=True, alias_generator=to_camel):
     """Default properties for binding constraints
--- a/src/antares/model/binding_constraint.py
+++ b/src/antares/model/binding_constraint.py
@@ -92,17 +92,7 @@ class ConstraintTerm(TermOperators):
         return ".".join((data.area.lower(), data.cluster.lower()))
 
     def serialize_term_data(self) -> Optional[str]:
-        """
-        Serializes the term data to be correctly written in INI.
-        """
-        if isinstance(self.data, LinkData):
-            if self.offset is not None:
-                return f"0.000000%{self.offset}"
-            if self.weight is not None:
-                return f"{self.weight}"
-            return "0"
-
-        return None
+        return self.weight_offset()
 
 
 class DefaultBindingConstraintProperties(BaseModel, extra="forbid", populate_by_name=True, alias_generator=to_camel):
--- a/src/antares/model/study.py
+++ b/src/antares/model/study.py
@@ -205,6 +205,7 @@ class Study:
         self._settings = DefaultStudySettings.model_validate(settings if settings is not None else StudySettings())
         self._areas: Dict[str, Area] = dict()
         self._links: Dict[str, Link] = dict()
+        self._binding_constraints: Dict[str, BindingConstraint] = dict()
 
     @property
     def service(self) -> BaseStudyService:
@@ -229,7 +230,7 @@ class Study:
         return self._settings
 
     def get_binding_constraints(self) -> MappingProxyType[str, BindingConstraint]:
-        return MappingProxyType(self._binding_constraints_service.binding_constraints)
+        return MappingProxyType(self._binding_constraints)
 
     def create_area(
         self, area_name: str, *, properties: Optional[AreaProperties] = None, ui: Optional[AreaUi] = None
@@ -269,9 +270,25 @@ class Study:
         equal_term_matrix: Optional[pd.DataFrame] = None,
         greater_term_matrix: Optional[pd.DataFrame] = None,
     ) -> BindingConstraint:
-        return self._binding_constraints_service.create_binding_constraint(
+        """
+        Create a new binding constraint and store it.
+
+        Args:
+            name (str): The name of the binding constraint.
+            properties (Optional[BindingConstraintProperties]): Optional properties for the constraint.
+            terms (Optional[List[ConstraintTerm]]): Optional list of terms for the constraint.
+            less_term_matrix (Optional[pd.DataFrame]): Optional less-than term matrix.
+            equal_term_matrix (Optional[pd.DataFrame]): Optional equality term matrix.
+            greater_term_matrix (Optional[pd.DataFrame]): Optional greater-than term matrix.
+
+        Returns:
+            BindingConstraint: The created binding constraint.
+        """
+        binding_constraint = self._binding_constraints_service.create_binding_constraint(
             name, properties, terms, less_term_matrix, equal_term_matrix, greater_term_matrix
         )
+        self._binding_constraints[binding_constraint.name] = binding_constraint
+        return binding_constraint
 
     def update_settings(self, settings: StudySettings) -> None:
         new_settings = self._study_service.update_study_settings(settings)
@@ -280,7 +297,7 @@ class Study:
 
     def delete_binding_constraint(self, constraint: BindingConstraint) -> None:
         self._study_service.delete_binding_constraint(constraint)
-        self._binding_constraints_service.binding_constraints.pop(constraint.id)
+        self._binding_constraints.pop(constraint.id)
 
     def delete(self, children: bool = False) -> None:
         self._study_service.delete(children)
--- a/src/antares/model/study.py
+++ b/src/antares/model/study.py
@@ -287,7 +287,7 @@ class Study:
         binding_constraint = self._binding_constraints_service.create_binding_constraint(
             name, properties, terms, less_term_matrix, equal_term_matrix, greater_term_matrix
         )
-        self._binding_constraints[binding_constraint.name] = binding_constraint
+        self._binding_constraints[binding_constraint.id] = binding_constraint
         return binding_constraint
 
     def update_settings(self, settings: StudySettings) -> None:
--- a/src/antares/service/local_services/binding_constraint_local.py
+++ b/src/antares/service/local_services/binding_constraint_local.py
@@ -9,7 +9,7 @@
 # SPDX-License-Identifier: MPL-2.0
 #
 # This file is part of the Antares project.
-from typing import Any, Optional
+from typing import Any, Optional, Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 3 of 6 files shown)


src/antares/model/study.py
```diff
@@ -246,9 +247,25 @@ def create_binding_constraint(
...
+            equal_term_matrix (Optional[pd.DataFrame]): Optional equality term matrix.
+            greater_term_matrix (Optional[pd.DataFrame]): Optional greater-than term matrix.
+
+        Returns:
+            BindingConstraint: The created binding constraint.
+        """
+        binding_constraint = self._binding_constraints_service.create_binding_constraint(
             name, properties, terms, less_term_matrix, equal_term_matrix, greater_term_matrix
         )
+        self._binding_constraints[binding_constraint.name] = binding_constraint
```
MartinBelthle, line 267: We usually use the ids as the key instead of the name

src/antares/service/local_services/binding_constraint_local.py
```diff
@@ -103,20 +95,76 @@ def _check_if_empty_ts(time_step: BindingConstraintFrequency, time_series: Optio
...
+            constraint_name (str): The name of the constraint.
+            constraint_id (str): The ID of the constraint.
+            terms (dict[str, ConstraintTerm], optional): Terms applying to the binding constraint. Defaults to None.
+        Raises:
+        BindingConstraintCreationError: If a binding constraint with the same name already exists in the INI file.
+        """
+
+        current_ini_content = self.ini_file.ini_dict or {}
+
+        if constraint_name in current_ini_content:
```
MartinBelthle, line 119: It's a bit more complicated as we want to ensure we don't create 2 same **ids** as there will be a conflict inside the filesystem. Example: 2 BCs; bc_1 and BC_1 don't have the same name but have the same ids and therefore cannot co-exist on the fs so we have to raise an Exception in such case

To calculate the id without having to infer the way it's done we should probably introduce a static method inside a Constraint Object that will return the transform function. This way if we one day change the way a BC id is calculated the code will still work

src/antares/service/local_services/binding_constraint_local.py
```diff
@@ -103,20 +95,76 @@ def _check_if_empty_ts(time_step: BindingConstraintFrequency, time_series: Optio
...
+            if terms in [None, {}]:
```
MartinBelthle, line 120: how can we have terms = {} as you typed it as a list ?
Also Pycharm tells me to merge the 2 if statements 
And also I don't get why you check the terms to know if you raise or not ? It seems it's because it one case you're updating and in the other creating. If so, it should be more explicit or separated in 2 methods

tests/antares/services/local_services/test_study.py
```diff
@@ -1498,7 +1498,7 @@ def test_creating_duplicate_area_name_errors(self, local_study_w_areas):
         # Then
         with pytest.raises(
             AreaCreationError,
-            match=f"Could not create the area {area_to_create}: There is already an area '{area_to_create}' in the study '{local_study_w_areas.name}'",
+            match=f"Could not create the area {area_to_create}: There is already an area '{area_to_create}' in the study '{local_study_w_areas.name}",
```
MartinBelthle, line 1501: This shouldn't change i think as you have only one ` at the beginning but not in the end

tests/antares/services/local_services/test_study.py
```diff
@@ -2099,7 +2099,7 @@ def test_constraints_ini_have_correct_default_content(
         self, local_study_with_constraint, test_constraint, default_constraint_properties
     ):
         # Given
-        expected_ini_contents = """[0]
+        expected_ini_contents = """[test constraint]
```
MartinBelthle, line 2102: Actually we should have 0 here it's the weird convention of this ini file, the section is the number of the constraint

Same goes for the other tests you modified

src/antares/service/local_services/binding_constraint_local.py
```diff
@@ -103,20 +95,76 @@ def _check_if_empty_ts(time_step: BindingConstraintFrequency, time_series: Optio
...
+
+        new_terms = {
+            **constraint.local_properties.terms,  # Existing terms
+            **{term.id: term for term in terms if term.id not in constraint.get_terms()},  # New terms
         }
+
         constraint.local_properties.terms = new_terms
-        self._write_binding_constraint_ini()
+
+        list(new_terms.values())
```
MartinBelthle, line 159: Typo: Should be removed

src/antares/service/local_services/binding_constraint_local.py
```diff
@@ -103,20 +95,76 @@ def _check_if_empty_ts(time_step: BindingConstraintFrequency, time_series: Optio
...
+
+        self._write_binding_constraint_ini(
```
MartinBelthle, line 161: Minor perf improvement: you could store list(new_terms.values()) inside a variable to avoid doing this twice (lines 165 and 168)

MartinBelthle (changes requested):

tests/antares/services/local_services/test_study.py
vargastat: Yes didnt know about this specifity for bc

src/antares/service/local_services/binding_constraint_local.py
vargastat: done

src/antares/tools/ini_tool.py
```diff
@@ -132,6 +145,9 @@ def update_from_ini_file(self) -> None:
 
         self._ini_contents = parsed_ini
 
+    def update_binding_constraints_from_ini_file(self) -> None:
```
MartinBelthle, line 148: You don't use this method, you can remove it

src/antares/service/local_services/binding_constraint_local.py
```diff
@@ -22,22 +21,41 @@
...
     ConstraintTerm,
+    LinkData,
 )
 from antares.service.base_services import BaseBindingConstraintService
 from antares.tools.ini_tool import IniFile, IniFileTypes
 from antares.tools.matrix_tool import df_save
 from antares.tools.time_series_tool import TimeSeriesFileType
 
 
+def serialize_term_data(
```
MartinBelthle, line 36: nice to add this utility method but I believe you should just have the term as an argument and treat offset and wieght internally as they are class variables.

In fact I think this should be a method of the class ConstraintTerm. It would be even better

Also i think we have the same writing for both terms so I don't understand the if else return None

src/antares/service/local_services/binding_constraint_local.py
```diff
@@ -54,17 +72,16 @@ def create_binding_constraint(
             properties=properties,
             terms=terms,
         )
-        if constraint.id in self.binding_constraints:
+        constraint.properties = constraint.local_properties.yield_binding_constraint_properties()
+
+        current_ini_content = self.ini_file.ini_dict_binding_constraints or {}
+        if any(values.get("name") == name for values in current_ini_content.values()):
```
MartinBelthle, line 78: As I said in my first review we should check ids and not name

src/antares/service/local_services/binding_constraint_local.py
MartinBelthle: You can change line 208 to return the introduced variables terms_values

src/antares/service/local_services/binding_constraint_local.py
```diff
@@ -103,20 +120,91 @@ def _check_if_empty_ts(time_step: BindingConstraintFrequency, time_series: Optio
...
+            constraint (BindingConstraint): The binding constraint to update.
+            terms (list[ConstraintTerm]): A list of new terms to add.
+
+        Returns:
+            list[ConstraintTerm]: The updated list of terms.
+        """
+
+        new_terms = {
+            **constraint.local_properties.terms,
+            **{term.id: term for term in terms if term.id not in constraint.get_terms()},
```
MartinBelthle, line 194: Shouldn't we raise if we want to add an already existing term instead of just not writing it ? Don't know

src/antares/service/local_services/binding_constraint_local.py
```diff
@@ -103,20 +120,91 @@ def _check_if_empty_ts(time_step: BindingConstraintFrequency, time_series: Optio
...
+
+            serialized_terms = (
+                {term.id: serialize_term_data(term.data, term.offset, term.weight) for term in terms} if terms else {}
+            )
+
+            existing_terms.update(serialized_terms)  # type: ignore
+            current_ini_content[existing_section] = existing_terms
+
+            # Persist the updated INI content
+            self.ini_file.ini_dict_binding_constraints = current_ini_content
```
MartinBelthle, line 153: I think you don't have to do this as you're doing the same thing line 177 no matter the if / else

src/antares/service/local_services/binding_constraint_local.py
```diff
@@ -103,20 +120,91 @@ def _check_if_empty_ts(time_step: BindingConstraintFrequency, time_series: Optio
...
+            self.ini_file.write_ini_file()
+        else:
+            terms_dict = (
+                {
+                    term.id: ConstraintTerm(data=term.data, offset=term.offset, weight=term.weight)
+                    if isinstance(term.data, (LinkData, ClusterData))
```
MartinBelthle, line 159: Don't think you need this if else as these are the 2 only possibilities for term.data

src/antares/service/local_services/binding_constraint_local.py
```diff
@@ -103,20 +120,91 @@ def _check_if_empty_ts(time_step: BindingConstraintFrequency, time_series: Optio
...
+            existing_terms.update(serialized_terms)  # type: ignore
+            current_ini_content[existing_section] = existing_terms
+
+            # Persist the updated INI content
+            self.ini_file.ini_dict_binding_constraints = current_ini_content
+            self.ini_file.write_ini_file()
+        else:
+            terms_dict = (
+                {
+                    term.id: ConstraintTerm(data=term.data, offset=term.offset, weight=term.weight)
```
MartinBelthle, line 158: Could be term.id: term no ?

src/antares/tools/ini_tool.py
```diff
@@ -90,6 +90,19 @@ def ini_dict(self, new_ini_dict: dict[str, dict[str, str]]) -> None:
...
 
+    @property
+    def ini_dict_binding_constraints(self) -> dict[str, dict[str, str]]:
+        return {section: dict(self._ini_contents[section]) for section in self._ini_contents.sections()}
+
+    @ini_dict_binding_constraints.setter
+    def ini_dict_binding_constraints(self, new_ini_dict: dict[str, dict[str, str]]) -> None:
+        """Set INI file contents for binding constraints."""
+        self._ini_contents = CustomRawConfigParser()
+        for index, (section, values) in enumerate(new_ini_dict.items()):
```
MartinBelthle, line 101: Could be for index, values in enumerate(new_ini_dict.values()) I think as you don't use section

MartinBelthle (changes requested):

src/antares/service/local_services/binding_constraint_local.py
vargastat: yes! done

src/antares/tools/ini_tool.py
vargastat: you are right!

src/antares/service/local_services/binding_constraint_local.py
vargastat: exact!

src/antares/model/binding_constraint.py
```diff
@@ -91,6 +91,19 @@ def generate_id(cls, data: Union[Dict[str, str], LinkData, ClusterData]) -> str:
...
+        Serializes the term data to be correctly written in INI.
+        """
+        if isinstance(self.data, LinkData):
+            if self.offset is not None:
+                return f"0.000000%{self.offset}"
+            if self.weight is not None:
+                return f"{self.weight}"
+            return "0"
+
+        return None
```
MartinBelthle, line 105: I just realized there's already a method doing this who don't have the issues I told in a previous reivew so you should use it : `weight_offset` inside TermOperators class

src/antares/service/local_services/binding_constraint_local.py
MartinBelthle: Not resolved :/. We should check the ids and for that introduce a static method inside the class BindingConstraint to get the method use to calculate an id based on a name

src/antares/service/local_services/binding_constraint_local.py
MartinBelthle: Not resolved :/. I think we should raise

MartinBelthle (changes requested):

src/antares/model/binding_constraint.py
```diff
@@ -91,6 +91,9 @@ def generate_id(cls, data: Union[Dict[str, str], LinkData, ClusterData]) -> str:
             return "%".join(sorted((data.area1.lower(), data.area2.lower())))
         return ".".join((data.area.lower(), data.cluster.lower()))
 
+    def serialize_term_data(self) -> Optional[str]:
```
MartinBelthle, line 94: Is it really useful why don't we directly call weight_offset() ?

src/antares/service/local_services/binding_constraint_local.py
```diff
@@ -103,21 +101,84 @@ def _check_if_empty_ts(time_step: BindingConstraintFrequency, time_series: Optio
...
+
+        Returns:
+            list[ConstraintTerm]: The updated list of terms.
+        """
+
+        new_terms = constraint.local_properties.terms.copy()
+
+        for term in terms:
+            if term.id in constraint.get_terms():
+                raise BindingConstraintCreationError(
```
MartinBelthle, line 165: Minor: the exception class name doesn't really fit as we're trying to add terms

MartinBelthle (commented):
Minor reviews

PR #26, opened 2024-12-05
feat(api): add create_variant methods

salemsd:
Added create_variant method and tests for api service. (Raise exception for local)


src/antares/service/local_services/study_local.py
```diff
@@ -40,3 +40,6 @@ def delete_binding_constraint(self, constraint: BindingConstraint) -> None:
 
     def delete(self, children: bool) -> None:
         raise NotImplementedError
+
+    def create_variant(self, variant_name: str) -> str:
```
MartinBelthle, line 44: Should return a Study more than a str
Same remark for the base method and the api implem

src/antares/service/api_services/study_api.py
```diff
@@ -104,3 +105,11 @@ def delete(self, children: bool) -> None:
             self._wrapper.delete(url)
         except APIError as e:
             raise StudyDeletionError(self.study_id, e.message) from e
+
+    def create_variant(self, variant_name: str) -> str:
+        url = f"{self._base_url}/studies/{self.study_id}/variants?name={variant_name}"
+        try:
+            response = self._wrapper.post(url)
+            return response.text
```
MartinBelthle, line 113: Adapt the code to return a Study.
You should do the same thing you did in the other create_variant function you introduced (but this time using read_areas())

src/antares/model/study.py
```diff
@@ -285,6 +302,24 @@ def delete_binding_constraint(self, constraint: BindingConstraint) -> None:
...
+        Creates a new variant for the study
+
+        Args:
+            variant_name: the name of the new variant
+        Returns: The variant in the form of a Study object
+        """
+        variant_id = self._study_service.create_variant(variant_name)
+        config = self._study_service.config
+
+        if isinstance(config, APIconf):
```
MartinBelthle, line 316: All of this should be handled inside the API layer inside the create_variant method as we don't want to put API or local related things inside the study model

tests/antares/services/api_services/test_study_api.py
```diff
@@ -250,3 +251,70 @@ def test_read_study_api(self):
...
             assert actual_study.version == expected_study.version
             assert actual_study.service.study_id == expected_study.service.study_id
+
+    def test_create_variant_success(self):
+        variant_name = "variant_test"
+        with requests_mock.Mocker() as mocker:
+            base_url = "https://antares.com/api/v1"
+            url = f"{base_url}/studies/{self.study_id}/variants?name={variant_name}"
+            variant_id = "variant_id"
+            mocker.post(url, text=variant_id, status_code=201)
```
MartinBelthle, line 261: You should use res.json() to get the study_id so you should mock json instead of text here

tests/antares/services/api_services/test_study_api.py
```diff
@@ -250,3 +251,70 @@ def test_read_study_api(self):
...
+            areas_url = f"{base_url}/studies/{variant_id}/areas?ui=true"
+            mocker.get(areas_url, json={}, status_code=200)
+
+            variant = self.study.create_variant(variant_name)
+
+            assert isinstance(variant, Study)
+            assert variant.name == variant_name
+            assert variant.service.study_id == variant_id
+
+    def test_create_variant_api_success(self):
```
MartinBelthle, line 278: As it's the exact same code except for the part where you use the method, i think it would be easier to understand to put all inside one test to avoid duplicating code

tests/antares/services/api_services/test_study_api.py
```diff
@@ -250,3 +251,70 @@ def test_read_study_api(self):
...
+        with requests_mock.Mocker() as mocker:
+            base_url = "https://antares.com/api/v1"
+            url = f"{base_url}/studies/{self.study_id}/variants?name={variant_name}"
+            error_message = "Variant creation failed"
+            mocker.post(url, json={"description": error_message}, status_code=404)
+
+            with pytest.raises(StudyVariantCreationError, match=error_message):
+                self.study.create_variant(variant_name)
+
+    def test_create_variant_api_fails(self):
```
MartinBelthle, line 312: Same remark here

MartinBelthle (commented):
Pretty nice :) few comments and also it would be nice to add an integration test inside test_web_client.py to ensure the code works

tests/integration/test_web_client.py
```diff
@@ -462,3 +462,17 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop):
         empty_settings = StudySettings()
         new_study.update_settings(empty_settings)
         assert old_settings == new_study.get_settings()
+
+    def test_create_variant(self, antares_web: AntaresWebDesktop):
```
MartinBelthle, line 466: Using the fixture antares_web launches the Desktop version which can take a long time. I would prefer you put this code inside the other test. Otherwise the test is nice (you could add the same test on the other creation method you introduced)

src/antares/service/local_services/study_local.py
```diff
@@ -10,13 +10,16 @@
...
 
-from typing import Any, Optional
+from typing import TYPE_CHECKING, Any, Optional
 
 from antares.config.local_configuration import LocalConfiguration
 from antares.model.binding_constraint import BindingConstraint
 from antares.model.settings.study_settings import StudySettings
 from antares.service.base_services import BaseStudyService
 
+if TYPE_CHECKING:
```
MartinBelthle, line 20: Why did you add this ? Was it to avoid circular imports ?

MartinBelthle (changes requested):
1 question and 1 change otherwise it's nice :)

src/antares/service/local_services/study_local.py
salemsd: Yes, that was to avoid circular imports with Study and the study services importing each other

tests/integration/test_web_client.py
```diff
@@ -462,3 +462,16 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop):
         empty_settings = StudySettings()
         new_study.update_settings(empty_settings)
         assert old_settings == new_study.get_settings()
+
+        # tests variant creation
+        parent_study = create_study_api("parent-test", "880", api_config)
```
MartinBelthle, line 467: Last comment: It would be better to test with an existing study instead of having to create another one just for this test.
Also you could use the create_variant_api function to create a 2nd variant and perform the same tests

MartinBelthle (changes requested):

MartinBelthle (approved):

PR #27, opened 2024-12-06
[REFACTO] Link created using area names instead of Area objects

vargastat:


src/antares/service/api_services/link_api.py
```diff
@@ -60,7 +60,9 @@ def create_link(
         """
         base_url = f"{self._base_url}/studies/{self.study_id}"
         # TODO: Currently, AntaresWeb does not have a specific endpoint for links. Once it will, we should change this logic.
-        area1_id, area2_id = sorted([area_from.id, area_to.id])
+
+        area1_id, area2_id = sorted([area_from, area_to])
```
MartinBelthle, line 64: We'll have the issue everywhere in the code. It could be nice to do this sort inside the Link object inside the `__init__`

You can remove this everywhere else if you do this

src/antares/service/local_services/link_local.py
```diff
@@ -33,8 +33,8 @@ def __init__(self, config: LocalConfiguration, study_name: str, **kwargs: Any) -
 
     def create_link(
         self,
-        area_from: Area,
-        area_to: Area,
+        area_from: str,
+        area_to: str,
         properties: Optional[LinkProperties] = None,
         ui: Optional[LinkUi] = None,
         existing_areas: Optional[MappingProxyType[str, Area]] = None,
```
MartinBelthle, line 40: I didn't see Sigurd introduced this to ensure validation inside local. Not a huge fan at all. We should do this inside the implem and remove this. But we can do this inside another PR

MartinBelthle (changes requested):
1 comment + another for this PR if you want to otherwise in another PR
Also rename your PR as it's not really clear IMO

src/antares/service/api_services/link_api.py
vargastat: Removed

src/antares/service/local_services/link_local.py
vargastat: PR to come

PR #28, opened 2024-12-06
refactor(link): parameter `existing_areas` removed from `create_link` method

vargastat:


src/antares/model/study.py
```diff
@@ -264,9 +264,14 @@ def create_link(
         area_to: str,
         properties: Optional[LinkProperties] = None,
         ui: Optional[LinkUi] = None,
-        existing_areas: Optional[MappingProxyType[str, Area]] = None,
     ) -> Link:
-        link = self._link_service.create_link(area_from, area_to, properties, ui, existing_areas)
+        missing_areas = [
+            area for area in [area_from, area_to] if area not in self._areas or not hasattr(self._areas[area], 'id')
+        ]
+        if missing_areas:
```
MartinBelthle, line 271: We should perform 2 checks:
- The one you did
- Asserts that there's not the same link already created in the study.

Also, you can remove the check with the `hasattr` as it doesn't really make sense IMO

MartinBelthle (changes requested):
You can add the little check + a test and we're good to go

src/antares/service/local_services/link_local.py
```diff
@@ -55,13 +52,6 @@ def create_link(
         """
         areas = dict(sorted({area_from: area_from, area_to: area_to}.items()))
```
MartinBelthle, line 53: Minor: Line 53 and 55 can now be simplifed in `area_from, area_to =  sorted([area_from, area_to])`

tests/antares/services/api_services/test_study_api.py
```diff
@@ -268,3 +287,48 @@ def test_create_variant_fails(self):
...
+            Mock(),
+        )
+
+        self.study._links[f"{area_from}/{area_to}"] = Link(area_from, area_to, Mock())
+
+        with pytest.raises(
+            LinkCreationError,
+            match=f"Could not create the link {area_from} / {area_to}: A link from {area_from} to {area_to} already exists",
+        ):
+            self.study.create_link(area_from="area_1", area_to="area_2")
```
MartinBelthle, line 316: Could be `area_from=area_from, area_to=area_to` to simplify readability

src/antares/model/study.py
```diff
@@ -264,9 +264,18 @@ def create_link(
...
         properties: Optional[LinkProperties] = None,
         ui: Optional[LinkUi] = None,
-        existing_areas: Optional[MappingProxyType[str, Area]] = None,
     ) -> Link:
-        link = self._link_service.create_link(area_from, area_to, properties, ui, existing_areas)
+        missing_areas = [area for area in [area_from, area_to] if area not in self._areas]
+        if missing_areas:
+            raise LinkCreationError(area_from, area_to, f"{', '.join(missing_areas)} does not exist")
+
+        existing_link = next(
```
MartinBelthle, line 272: We should sort the areas to perform the check as we cannot have 2 links a -> b and b -> a.

Also, I just realized that I think we handle the MAJ badly. As we don't force lower case inside the property name inside the link object :/ + we use these values as if they were ids even they aren't

I think we should introduce a static method that returns the 2 names transformed in ids to use them internally. With this we could use this method here to check if the link already exists and this time we'll be sure the check works.

Could be done inside a next PR

MartinBelthle (changes requested):
Minor review + a change to perform in this PR or in a next one

PR #29, opened 2024-12-09
feat(local): add Thermal reading

killian-scalian:
- Thermal reading
-  Unitary test
- Co2Cost, fuelCost, data, modulation and series timeseries handled

### DIFF
--- a/src/antares/service/local_services/thermal_local.py
+++ b/src/antares/service/local_services/thermal_local.py
@@ -47,7 +47,7 @@ class ThermalLocalService(BaseThermalService):
             time_serie_type = TimeSeriesFileType.THERMAL_DATA
         elif ts_name.value == "CO2Cost":
             time_serie_type = TimeSeriesFileType.THERMAL_CO2
-        elif ts_name.value == "fuelCost":
+        else:
             time_serie_type = TimeSeriesFileType.THERMAL_FUEL
 
         return read_timeseries(
@@ -62,6 +62,7 @@ class ThermalLocalService(BaseThermalService):
         thermal_clusters = []
         if thermal_dict:
             for thermal_cluster in thermal_dict:
+                # TODO refactor this as it is really not clean
                 thermal_properties = ThermalClusterPropertiesLocal(
                     group=thermal_dict[thermal_cluster]["group"],
                     thermal_name=thermal_dict[thermal_cluster]["name"],
@@ -101,12 +102,12 @@ class ThermalLocalService(BaseThermalService):
                     variable_o_m_cost=thermal_dict[thermal_cluster]["variableomcost"],
                 )
 
-            thermal_clusters.append(
-                ThermalCluster(
-                    thermal_service=self,
-                    area_id=area_id,
-                    name=thermal_dict[thermal_cluster]["name"],
-                    properties=thermal_properties.yield_thermal_cluster_properties(),
+                thermal_clusters.append(
+                    ThermalCluster(
+                        thermal_service=self,
+                        area_id=area_id,
+                        name=thermal_dict[thermal_cluster]["name"],
+                        properties=thermal_properties.yield_thermal_cluster_properties(),
+                    )
                 )
-            )
         return thermal_clusters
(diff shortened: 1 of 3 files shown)


tests/antares/services/local_services/test_area.py
```diff
@@ -1306,3 +1307,76 @@ def test_read_misc_gen_local(self, local_study_w_areas):
             _write_file(file_path, expected_time_serie)
             matrix = area.get_misc_gen_matrix()
             pd.testing.assert_frame_equal(matrix, expected_time_serie)
+
+
+class TestReadThermal:
+    def create_file_with_empty_dataframe(self, file_path: str, df: pd.DataFrame):
```
MartinBelthle, line 1313: You don't use this method, you can remove it

src/antares/service/local_services/thermal_local.py
```diff
@@ -31,7 +39,74 @@ def update_thermal_properties(
...
+                    op2=thermal_dict[thermal_cluster]["op2"],
+                    op3=thermal_dict[thermal_cluster]["op3"],
+                    op4=thermal_dict[thermal_cluster]["op4"],
+                    op5=thermal_dict[thermal_cluster]["op5"],
+                    cost_generation=thermal_dict[thermal_cluster]["costgeneration"],
+                    efficiency=thermal_dict[thermal_cluster]["efficiency"],
+                    variable_o_m_cost=thermal_dict[thermal_cluster]["variableomcost"],
+                )
+
+            thermal_clusters.append(
```
MartinBelthle, line 104: Put this inside the loop, it's not indented correctly

src/antares/service/local_services/thermal_local.py
```diff
@@ -31,7 +39,74 @@ def update_thermal_properties(
...
+            cluster_id=thermal_cluster.properties.group.value,
+        )
 
     def read_thermal_clusters(self, area_id: str) -> List[ThermalCluster]:
-        raise NotImplementedError
+        thermal_dict = IniFile(self.config.study_path, IniFileTypes.THERMAL_LIST_INI, area_name=area_id).ini_dict
+        thermal_clusters = []
+        if thermal_dict:
+            for thermal_cluster in thermal_dict:
+                thermal_properties = ThermalClusterPropertiesLocal(
```
MartinBelthle, line 65: This is really really ugly. We should introduce aliases inside the `ThermalClusterProperitesLocal` so inside the `DefaultThermalProperties` class to use pydantic to validate this. And this goes with adding the aliases camel case inside the `API` class.
But this is too complicated for this PR so you can just add a `TODO` saying that we should refactor this

src/antares/service/local_services/thermal_local.py
```diff
@@ -31,7 +39,74 @@ def update_thermal_properties(
...
+            time_serie_type = TimeSeriesFileType.THERMAL_SERIES
+        elif ts_name.value == "modulation":
+            time_serie_type = TimeSeriesFileType.THERMAL_MODULATION
+        elif ts_name.value == "data":
+            time_serie_type = TimeSeriesFileType.THERMAL_DATA
+        elif ts_name.value == "CO2Cost":
+            time_serie_type = TimeSeriesFileType.THERMAL_CO2
+        elif ts_name.value == "fuelCost":
+            time_serie_type = TimeSeriesFileType.THERMAL_FUEL
+
```
MartinBelthle, line 52: You should either put a else: raise NotImplementedError here or a else with the fuelCost as here we can reach a null pointer as `time_serie_type` could be undefined

MartinBelthle (changes requested):

src/antares/service/local_services/thermal_local.py
killian-scalian: Ok, for the last commit

src/antares/service/local_services/thermal_local.py
killian-scalian: Ok, for the last commit

tests/antares/services/local_services/test_area.py
killian-scalian: Ok, for the last commit

PR #30, opened 2024-12-09
feat(local): add Links reading

killian-scalian:
- Add reading part for the links
-  to the read_timeseries methode, add one optional input
- Add unitary test

### DIFF
--- a/src/antares/model/link.py
+++ b/src/antares/model/link.py
@@ -15,7 +15,7 @@ from typing import Mapping, Optional, Set
 
 import pandas as pd
 
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 from antares.model.commons import FilterOption, sort_filter_values
 from antares.tools.alias_generators import to_kebab
@@ -76,6 +76,15 @@ class LinkProperties(DefaultLinkProperties):
 
 
 class LinkPropertiesLocal(DefaultLinkProperties):
+    hurdles_cost: bool = Field(default=False, alias="hurdles-cost")
+    loop_flow: bool = Field(default=False, alias="loop-flow")
+    use_phase_shifter: bool = Field(default=False, alias="use-phase-shifter")
+    transmission_capacities: TransmissionCapacities = Field(
+        default=TransmissionCapacities.ENABLED, alias="transmission-capacities"
+    )
+    asset_type: AssetType = Field(default=AssetType.AC, alias="asset-type")
+    display_comments: bool = Field(default=True, alias="display-comments")
+
     @property
     def ini_fields(self) -> Mapping[str, str]:
         return {
@@ -114,6 +123,12 @@ class LinkUi(DefaultLinkUi):
 
 
 class LinkUiLocal(DefaultLinkUi):
+    link_style: LinkStyle = Field(default=LinkStyle.PLAIN, alias="link-style")
+    link_width: float = Field(default=1, alias="link-width")
+    colorr: int = Field(default=112, alias="colorr")
+    colorg: int = Field(default=112, alias="colorg")
+    colorb: int = Field(default=112, alias="colorb")
+
     @property
     def ini_fields(self) -> Mapping[str, str]:
         # todo: can be replaced with alias i believe
--- a/src/antares/model/link.py
+++ b/src/antares/model/link.py
@@ -76,14 +76,6 @@ class LinkProperties(DefaultLinkProperties):
 
 
 class LinkPropertiesLocal(DefaultLinkProperties):
-    hurdles_cost: bool = Field(default=False, alias="hurdles-cost")
-    loop_flow: bool = Field(default=False, alias="loop-flow")
-    use_phase_shifter: bool = Field(default=False, alias="use-phase-shifter")
-    transmission_capacities: TransmissionCapacities = Field(
-        default=TransmissionCapacities.ENABLED, alias="transmission-capacities"
-    )
-    asset_type: AssetType = Field(default=AssetType.AC, alias="asset-type")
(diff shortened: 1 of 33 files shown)


src/antares/service/local_services/link_local.py
```diff
@@ -133,3 +137,56 @@ def sort_link_properties_dict(ini_dict: Dict[str, str]) -> Dict[str, str]:
             "filter-year-by-year",
         ]
         return dict(sorted(ini_dict.items(), key=lambda item: dict_order.index(item[0])))
+
+    def get_capacity_direct(
```
MartinBelthle, line 141: You should add these 3 methods inside the base service and inside the API one

src/antares/service/local_services/link_local.py
```diff
@@ -133,3 +137,56 @@ def sort_link_properties_dict(ini_dict: Dict[str, str]) -> Dict[str, str]:
...
+        self,
+        area_id: str,
```
MartinBelthle, line 143: Also it should be typed area_from, area_to to fit with the existing naming we chose

MartinBelthle (changes requested):
Minor change requested

src/antares/service/local_services/link_local.py
```diff
@@ -133,3 +137,56 @@ def sort_link_properties_dict(ini_dict: Dict[str, str]) -> Dict[str, str]:
...
+        link_clusters = []
+        for element in link_path.iterdir():
+            area_from = element.name
+            for content_area in element.iterdir():
+                if (
+                    content_area.is_file()
+                    and content_area.suffix == ".txt"
+                    and content_area.stem.endswith("parameters")
+                ):
+                    link_clusters.append(
```
MartinBelthle, line 185: You didn't do the part on LinkUi and LinkProperties. Not really a big deal if you add a TODO but you can do this inside this PR

MartinBelthle (changes requested):

src/antares/service/local_services/link_local.py
killian-scalian: Fixed in the last commit

src/antares/service/local_services/link_local.py
killian-scalian: I did it with the aliases validation

src/antares/model/link.py
```diff
@@ -112,6 +123,12 @@ class LinkUi(DefaultLinkUi):
 
 
 class LinkUiLocal(DefaultLinkUi):
+    link_style: LinkStyle = Field(default=LinkStyle.PLAIN, alias="link-style")
+    link_width: float = Field(default=1, alias="link-width")
+    colorr: int = Field(default=112, alias="colorr")
```
MartinBelthle, line 128: Don't need to introduce aliases when it's the same as the python variable

src/antares/model/link.py
```diff
@@ -74,6 +76,15 @@ class LinkProperties(DefaultLinkProperties):
 
 
 class LinkPropertiesLocal(DefaultLinkProperties):
+    hurdles_cost: bool = Field(default=False, alias="hurdles-cost")
```
MartinBelthle, line 79: Don't think you should introduce these variables as they are already defined in the DefaultLinkProperties. And you don't use the aliases

MartinBelthle (changes requested):
2 minor changes

src/antares/model/link.py
killian-scalian: i removed all these alias. I thought it was useful but no changes when i remove it

src/antares/model/link.py
killian-scalian: Removed next commit

tests/antares/services/local_services/test_area.py
```diff
@@ -1374,3 +1375,57 @@ def test_read_thermals_local(self, local_study_w_thermal):
...
+        for link in links:
+            if link.area_from == "fr":
+                matrix = link.get_capacity_direct()
+                matrix_2 = link.get_capacity_indirect()
+                matrix_3 = link.get_parameters()
+
+                pd.testing.assert_frame_equal(matrix.astype(str), df_initial.astype(str), check_dtype=False)
+                pd.testing.assert_frame_equal(matrix_2.astype(str), df_initial.astype(str), check_dtype=False)
+                pd.testing.assert_frame_equal(matrix_3.astype(str), df_initial.astype(str), check_dtype=False)
+            assert link.ui.link_style.value == "plain"
```
Sigurd-Borge, line 1420: Here you're simply testing the default values for each link, not a big deal but doesn't seem necessary to test for all links unless you think there's a random factor that impacts the writing of the files? The default values already have testing in the creation of links.

killian-scalian: Here the dataframes are created inside the unitary test. So i compare what i read with is written at the start. But indeed it is not very useful to assert all these 3 matrix

Sigurd-Borge: I meant these:
```python
            assert link.ui.link_style.value == "plain"
            assert link.ui.link_width == 1
            assert link.ui.colorb == 112
            assert link.ui.colorg == 112
            assert link.ui.colorr == 112
            assert not link.properties.hurdles_cost
            assert not link.properties.loop_flow
            assert not link.properties.use_phase_shifter
            assert link.properties.transmission_capacities.value == "enabled"
            assert link.properties.asset_type.value == "ac"
            assert isinstance(link.properties.filter_year_by_year, set)
            assert isinstance(link.properties.filter_synthesis, set)
```

MartinBelthle (approved):
Okay for me, I don't really care about the test do as you like

PR #31, opened 2024-12-10
feat(local): add local integration test

Sigurd-Borge:


src/antares/exceptions/exceptions.py
```diff
@@ -54,6 +54,15 @@ def __init__(self, area_name: str, message: str) -> None:
         super().__init__(self.message)
 
 
+class InvalidNameError(Exception):
```
MartinBelthle, line 57: Name can contain these types of characters, it's the ids that cannot. What you should do is use transform_name_to_id(name) where you want to have an id but you shouldn't use this

So you'll have to adapt the rest of your code :/

tests/antares/services/local_services/test_study.py
```diff
@@ -1438,7 +1440,10 @@ def mock_error_in_sets_ini():
             raise CustomError("An error occurred while processing area can not be created")
 
         monkeypatch.setattr("antares.service.local_services.area_local._sets_ini_content", mock_error_in_sets_ini)
-        with pytest.raises(CustomError, match="An error occurred while processing area can not be created"):
+        with pytest.raises(
+            AreaCreationError,
+            match="Could not create the area test: Error during area creation: An error occurred while processing area can not be created",
```
MartinBelthle, line 1445: This is a really long msg, it should be shortened

tests/integration/conftest.py
```diff
@@ -0,0 +1,32 @@
...
+    study_version = "880"
+
+    # Study
+    other_study = create_study_local(other_study_name, study_version, tmp_path.absolute())
+
+    return other_study
+
+
+@pytest.fixture
+def other_area(other_study):
```
MartinBelthle, line 30: Find better fixture names, it's not really clear

tests/integration/test_local_client.py
```diff
@@ -0,0 +1,279 @@
...
+from antares.model.settings.study_settings import StudySettingsLocal
+from antares.model.st_storage import STStorageGroup, STStorageProperties
+from antares.model.study import Study
+from antares.model.thermal import ThermalCluster, ThermalClusterGroup, ThermalClusterProperties
+from antares.tools.ini_tool import IniFile, IniFileTypes
+from antares.tools.time_series_tool import TimeSeriesFileType
+
+
+class TestLocalClient:
+    def test_local_study(self, tmp_path, other_area):
```
MartinBelthle, line 41: Add a msg at the beginning of the test to explain what it's gonna perform

tests/integration/test_local_client.py
```diff
@@ -0,0 +1,279 @@
...
+
+        # Setup time series for following tests
+        time_series_rows = 10  # 365 * 24
+        time_series_columns = 1
+        time_series = pd.DataFrame(np.around(np.random.rand(time_series_rows, time_series_columns)))
+
+        # Load
+        fr.create_load(time_series)
+
+        assert test_study.service.config.study_path.joinpath(
```
MartinBelthle, line 85: You should test with the real path not the TimeSeriesFileType as if i write a wrong path inside the object the test will still pass and yet it will be wrong for the simulator

Same for the rest of the test

MartinBelthle (changes requested):

MartinBelthle:
Also, move your test inside the `antares` folder as otherwise it won't be tested inside the CI

src/antares/exceptions/exceptions.py
Sigurd-Borge: I've duplicated the functionality of AntaresWeb like asked, if that is no longer the case I wouldn't have made it this way.

image

I realize you can "force it" by using the API, but if for example you use a name with comma you can no longer correctly interact with your study using AntaresWeb (questionmark seems to work).

image

Which leads to:

image

tests/integration/conftest.py
Sigurd-Borge: Renamed

tests/integration/test_local_client.py
Sigurd-Borge: Updated

tests/antares/services/local_services/test_study.py
Sigurd-Borge: This is a combination of two levels of error messages raised in this particular test, as you see the two sets of colons
1) 'Could not create the area test:' this is built into AreaCreationError
2) 'Error during area create:' this is raised in create_area_local
3) The last bit is specific to this test

I have proceeded to delete 2 as I think it's superfluous.

tests/integration/test_local_client.py
Sigurd-Borge: Added.

Sigurd-Borge:
> Also, move your test inside the `antares` folder as otherwise it won't be tested inside the CI

Moved as requested.

src/antares/exceptions/exceptions.py
MartinBelthle: Indeed it's a bug inside AntaresWeb. But we shouldn't care about that inside antares-craft

MartinBelthle (changes requested):
I answered your comment it's the last comment

src/antares/exceptions/exceptions.py
Sigurd-Borge: Removed.

MartinBelthle (changes requested):
I'll review when the conflicts will be resolved :)

Sigurd-Borge:
Conflicts resolved:

tests/antares/services/local_services/test_study.py
```diff
@@ -1437,7 +1437,10 @@ def mock_error_in_sets_ini():
             raise CustomError("An error occurred while processing area can not be created")
 
         monkeypatch.setattr("antares.service.local_services.area_local._sets_ini_content", mock_error_in_sets_ini)
-        with pytest.raises(CustomError, match="An error occurred while processing area can not be created"):
+        with pytest.raises(
+            AreaCreationError,
+            match="Could not create the area test: An error occurred while processing area can not be created",
```
MartinBelthle, line 1442: Still not really clear, we don't have the area id and the msg isn't really in proper English

tests/antares/integration/conftest.py
```diff
@@ -0,0 +1,32 @@
...
+    study_version = "880"
+
+    # Study
+    other_study = create_study_local(other_study_name, study_version, tmp_path.absolute())
+
+    return other_study
+
+
+@pytest.fixture
+def unknown_area(other_study_with_other_area):
```
MartinBelthle, line 30: The fixture names still aren't clear IMO, it's a study with an area, not just an area

MartinBelthle (approved):
Some things are still weird but I don't really care as it concerns the local

PR #32, opened 2024-12-11
refactor(local): hide local implementation from constraints model

salemsd:
- Removed every local related code from binding_constraint.py model
- Put everything in binding_constraint_local.py with a few changes
- Removed a test making sure that updating properties also updates local ones


src/antares/service/local_services/binding_constraint_local.py
```diff
@@ -131,14 +187,7 @@ def _write_binding_constraint_ini(
...
         else:
-            terms_dict = {term.id: term for term in terms} if terms else {}
-
-            full_properties = BindingConstraintPropertiesLocal(
-                constraint_name=constraint_name,
-                constraint_id=constraint_id,
-                terms=terms_dict,
-                **properties.model_dump(),
-            )
+            full_properties = BindingConstraintPropertiesLocal(**local_properties.model_dump())
```
MartinBelthle, line 190: Perhaps I'm missing something but it looks like you don't have to introduce this object and you can just use local_properties as it is

MartinBelthle (changes requested):
Nice :) just a little comment

PR #33, opened 2024-12-11
fix(link): use area ids instead of names to avoid several issues

vargastat:


tests/antares/services/api_services/test_study_api.py
```diff
@@ -289,31 +269,32 @@ def test_create_variant_fails(self):
                 create_variant_api(self.api, self.study_id, variant_name)
 
     def test_create_duplicated_link(self):
```
MartinBelthle, line 271: We missed one thing, creating a link that starts from and goes to the same area.
We should add this check and a test about it

MartinBelthle (changes requested):

PR #34, opened 2024-12-11
feat(api): handle link matrices

mehdiwahada:


src/antares/model/area.py
```diff
@@ -286,66 +287,81 @@ def create_st_storage(self, st_storage_name: str, properties: Optional[STStorage
...
     def update_ui(self, ui: AreaUi) -> None:
-        new_ui = self._area_service.update_area_ui(self, ui)
+        new_ui = self._area_service.update_area_ui(self.id, ui)
         self._ui = new_ui
 
     def create_load(self, series: pd.DataFrame) -> None:
-        self._area_service.create_load(self, series=series)
+        try:
+            self._area_service.create_load(self.id, series)
+        except APIError as e:
```
MartinBelthle, line 339: I don't like the fact that we're introducing API related things inside the model. We should put this try catch inside the API Implementation

src/antares/model/study.py
```diff
@@ -256,9 +256,9 @@ def create_area(
         self._areas[area.id] = area
         return area
 
-    def delete_area(self, area: Area) -> None:
-        self._area_service.delete_area(area)
-        self._areas.pop(area.id)
+    def delete_area(self, area_id: str) -> None:
```
MartinBelthle, line 259: Here I think we should still have an Area as an argument and then pass the id. It would be clearer for the user I believe

src/antares/service/api_services/link_api.py
```diff
@@ -160,20 +163,57 @@ def update_link_ui(self, link: Link, ui: LinkUi) -> LinkUi:
...
 
         except APIError as e:
-            raise LinkUiUpdateError(link.name, e.message) from e
+            raise LinkUiUpdateError(link.id, e.message) from e
 
         return link_ui
 
-    def read_links(self) -> list[Link]:
-        raise NotImplementedError
+    def get_parameters(self, area_from_id: str, area_to_id: str) -> pd.DataFrame:
```
MartinBelthle, line 170: We should choose between area_from_id and area_from between the 2 methods. I think we should type area_from_id as it's clearer

src/antares/service/api_services/link_api.py
```diff
@@ -160,20 +163,57 @@ def update_link_ui(self, link: Link, ui: LinkUi) -> LinkUi:
...
+        try:
+            parameters_path = f"input/links/{area_from_id}/{area_to_id}_parameters"
+            matrix = get_matrix(self._base_url, self.study_id, self._wrapper, parameters_path)
+        except APIError as e:
+            raise LinkDownloadError(f"{area_from_id}/{area_to_id}", "parameters", e.message)
```
MartinBelthle, line 175: You should raise from e instead of giving the e.message. see other parts of the code

src/antares/service/base_services.py
```diff
@@ -361,20 +361,58 @@ def update_link_ui(self, link: Link, ui: LinkUi) -> LinkUi:
...
+    def get_parameters(self, area_from_id: str, area_to_id: str) -> pd.DataFrame:
+        """
+        Returns: link parameters
+        """
+        pass
+
+    @abstractmethod
+    def create_parameters(self, series: pd.DataFrame, area_from: str, area_to: str) -> None:
+        """
+        Args:
```
MartinBelthle, line 374: Either remove the args or fill them. I think you can remove them

tests/antares/services/api_services/test_link_api.py
```diff
@@ -99,6 +107,168 @@ def test_update_links_ui_fails(self):
...
+    def test_create_parameters_success(self):
+        with requests_mock.Mocker() as mocker:
+            raw_url = (
+                f"https://antares.com/api/v1/studies/{self.study_id}/raw?path="
+                f"input/links/{self.area_from.id}/{self.area_to.id}_parameters"
+            )
+
+            mocker.post(raw_url, status_code=200)
+
+            self.link.create_parameters(pd.DataFrame(data=np.ones((8760, 1))))
```
MartinBelthle, line 123: You could use self.matrix instead of introducing a dataframe each time

tests/integration/test_web_client.py
```diff
@@ -92,6 +92,14 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop):
         assert area_be.ui.x == area_ui.x
         assert area_be.ui.color_rgb == area_ui.color_rgb
 
+        # tests study reading method and comparing ids, name, areas and settings
```
MartinBelthle, line 95: You can remove this part of the test as it's already done line 231

tests/integration/test_web_client.py
```diff
@@ -234,6 +254,8 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop):
         assert list(expected_area_fr.get_st_storages()) == list(actual_area_fr.get_st_storages())
         assert study.get_settings() == actual_study.get_settings()
 
+        # checking the links between two areas
```
MartinBelthle, line 257: You can remove this comment as there's nothing after it

MartinBelthle (changes requested):

src/antares/service/base_services.py
```diff
@@ -361,20 +361,53 @@ def update_link_ui(self, link: Link, ui: LinkUi) -> LinkUi:
...
+        Returns: the direct capacity of a link
+        """
+        pass
+
+    @abstractmethod
+    def create_capacity_direct(self, series: pd.DataFrame, area_from: str, area_to: str) -> None:
+        """
+        Args:
+            area_to:
+            area_from:
```
MartinBelthle, line 391: Last comment about this unresolved

src/antares/service/api_services/area_api.py
```diff
@@ -479,71 +483,75 @@ def update_area_ui(self, area: Area, ui: AreaUi) -> AreaUi:
...
             ui_response = AreaUiResponse.model_validate(json_ui)
             area_ui = AreaUi.model_validate(ui_response.to_craft())
 
         except APIError as e:
-            raise AreaUiUpdateError(area.id, e.message) from e
+            raise AreaUiUpdateError(area_id, e.message) from e
 
         return area_ui
 
     def delete_area(self, area: Area) -> None:
```
MartinBelthle, line 495: The delete area should only get the area_id to be the same as the other methods. The delete inisde the sutdy.py gets an Area but calls the service with the id

src/antares/service/api_services/link_api.py
```diff
@@ -160,20 +163,57 @@ def update_link_ui(self, link: Link, ui: LinkUi) -> LinkUi:
...
 
         except APIError as e:
-            raise LinkUiUpdateError(link.name, e.message) from e
+            raise LinkUiUpdateError(link.id, e.message) from e
 
         return link_ui
 
-    def read_links(self) -> list[Link]:
-        raise NotImplementedError
+    def get_parameters(self, area_from: str, area_to_id: str) -> pd.DataFrame:
```
MartinBelthle, line 170: You forgot to rename area_to_id in area_to

src/antares/service/api_services/link_api.py
```diff
@@ -160,20 +163,57 @@ def update_link_ui(self, link: Link, ui: LinkUi) -> LinkUi:
...
+        try:
+            parameters_path = f"input/links/{area_from}/{area_to_id}_parameters"
+            matrix = get_matrix(self._base_url, self.study_id, self._wrapper, parameters_path)
+        except APIError as e:
+            raise LinkDownloadError(f"{area_from}/{area_to_id}", "parameters", e.message) from e
```
MartinBelthle, line 175: You can remove the e.message as now you raise from e. Also remove it inside the exception class

src/antares/service/api_services/area_api.py
```diff
@@ -353,37 +354,44 @@ def create_st_storage(
...
-
-    def create_wind(self, area: Area, series: pd.DataFrame) -> None:
-        series_path = f"input/wind/series/wind_{area.id}"
-        self._upload_series(area, series, series_path, "wind")
+    def create_wind(self, area_id: str, series: pd.DataFrame) -> None:
+        try:
+            series_path = f"input/wind/series/wind_{area_id}"
+            upload_series(self._base_url, self.study_id, self._wrapper, series, series_path)
+        except APIError as e:
+            raise MatrixUploadError(area_id, "wind", e.message)
```
MartinBelthle, line 373: you forgot to put back the `from e` at the end

src/antares/service/api_services/link_api.py
```diff
@@ -160,20 +163,57 @@ def update_link_ui(self, link: Link, ui: LinkUi) -> LinkUi:
...
         return link_ui
 
-    def read_links(self) -> list[Link]:
-        raise NotImplementedError
+    def get_parameters(self, area_from: str, area_to: str) -> pd.DataFrame:
+        try:
+            parameters_path = f"input/links/{area_from}/{area_to}_parameters"
+            matrix = get_matrix(self._base_url, self.study_id, self._wrapper, parameters_path)
+        except APIError as e:
+            raise LinkDownloadError(f"{area_from}/{area_to}", "parameters", e.message) from e
```
MartinBelthle, line 175: Last remark:
The exception class should expect area_form and to as parameter and do the area_from / area_to itself. This way a new user won't have to know how it's done to call the exception

MartinBelthle (changes requested):
Last review :)

PR #35, opened 2024-12-12
feat(api): add possibility to run study simulations

salemsd:
- Studies now have a run service that can be called to both start a simulation and wait for its job
- Unit tests are working


src/antares/service/api_services/run_api.py
```diff
@@ -0,0 +1,89 @@
...
+        url = f"{self._base_url}/launcher/jobs/{job_id}"
+        response = self._wrapper.get(url)
+        job_info = response.json()
+        launcher_params = json.loads(job_info["launcher_params"])
+        return Job(job_id, job_info["status"], launcher_params["auto_unzip"])
+
+    def wait_job_completion(self, job: Job, time_out: int) -> None:
+        start_time = time.time()
+        repeat_interval = 5
+        while job.status in (JobStatus.RUNNING.value, JobStatus.PENDING.value):
```
MartinBelthle, line 61: I don't really like this while as if you could enter this function with an already failed job.
So i would prefer a classic loop :)

src/antares/service/api_services/run_api.py
```diff
@@ -0,0 +1,89 @@
...
+            if time.time() - start_time > time_out:
+                raise SimulationTimeOutError(job.job_id, time_out)
+            time.sleep(repeat_interval)
+            updated_job = self._get_job_from_id(job.job_id)
+            job.status = updated_job.status
+            job.unzip_output = updated_job.unzip_output
```
MartinBelthle, line 67: this won't change, the only yhing you need to update is the status

src/antares/service/api_services/run_api.py
```diff
@@ -0,0 +1,89 @@
...
+
+        if job.unzip_output:
```
MartinBelthle, line 69: Once you go out of the loop your status could be success or failed.
If it's failed, you should raise an Exception to say the simulation failed.
Else you can continue your code

src/antares/service/api_services/run_api.py
```diff
@@ -0,0 +1,89 @@
...
+            self._unzip_output(self.study_id, ["UNARCHIVE"])
+
+        return None
+
+    def _unzip_output(self, ref_id: str, type: list[str]) -> None:
```
MartinBelthle, line 74: Perhaps my spec wasn't really clear as it's a complex subject but what you need to do in this method is to wait for the output to unzip itself. So you can rename this method.
Also, your payload should simply be 
```json
{
    "type": type,
    "ref_id": ref_id
}
```
otherwise you won't have any result (example: to_completion_date: 0 means you want all jobs that ended before date 0 which will return no jobs)

src/antares/service/api_services/run_api.py
```diff
@@ -0,0 +1,89 @@
...
+                "status": [],
+                "name": "string",
+                "type": type,
+                "ref_id": ref_id,
+                "from_creation_date_utc": 0,
+                "to_creation_date_utc": 0,
+                "from_completion_date_utc": 0,
+                "to_completion_date_utc": 0,
+            }
+            self._wrapper.post(url, json=payload)
```
MartinBelthle, line 87: Finally here you should repeat this call every 2 seconds for example, and check the output to see if the task ended successfully. I can reexplain this to you if you need

src/antares/exceptions/exceptions.py
```diff
@@ -274,3 +274,23 @@ class ConfigurationError(Exception):
...
+
+
+class AntaresSimulationRunningError(Exception):
+    def __init__(self, study_id: str, message: str) -> None:
+        self.message = f"Could not run the simulation for study {study_id}: " + message
+        super().__init__(self.message)
+
+
+class SimulationTimeOutError(Exception):
+    def __init__(self, job_id: str, time_out: int, message: str = "Error") -> None:
```
MartinBelthle, line 286: Don't really see the point of the 'message' argument here

tests/antares/services/api_services/test_study_api.py
```diff
@@ -330,3 +336,60 @@ def test_create_link_same_area(self):
...
+
+    def test_run_and_wait_antares_simulation(self):
+        parameters = AntaresSimulationParameters(solver=Solver.COIN, nb_cpu=2, unzip_output=True, presolve=False)
+        with requests_mock.Mocker() as mocker:
+            run_url = f"https://antares.com/api/v1/launcher/run/{self.study_id}"
+            job_id = "1234-g6z17"
+            mocker.post(run_url, json={"id": job_id}, status_code=200)
+
+            job_url = f"https://antares.com/api/v1/launcher/jobs/{job_id}"
+            response_list = [
```
MartinBelthle, line 348: I'm not sure to understand this. The output shouldn't be a list when calling this endpoint. We can discuss it later.

MartinBelthle (changes requested):
Nice work :) Once the comments will be resolved you could add an integration test it would be great

src/antares/model/settings/antares_simulation_parameters.py
```diff
@@ -0,0 +1,41 @@
...
+    @property
+    def other_options(self) -> str:
+        options = []
+        if self.presolve:
+            options.append("presolve")
+        if self.solver != Solver.SIRIUS:
+            options.append(self.solver.name.lower())
+        return " ".join(options)
+
+    def model_dump(self, *args: Any, **kwargs: Any) -> Dict[str, Any]:
```
MartinBelthle, line 36: We should rename this method `to_api()` instaed of model_dump because when we'll implement the local method we'll have to parse the object differently

src/antares/service/api_services/run_api.py
```diff
@@ -0,0 +1,106 @@
...
+            return self._get_job_from_id(job_id)
+        except APIError as e:
+            raise AntaresSimulationRunningError(self.study_id, e.message) from e
+
+    def _get_job_from_id(self, job_id: str) -> Job:
+        url = f"{self._base_url}/launcher/jobs/{job_id}"
+        response = self._wrapper.get(url)
+        job_info = response.json()
+        status = JobStatus.from_str(job_info["status"])
+        output_id = job_info["output_id"] if status == JobStatus.SUCCESS else None
```
MartinBelthle, line 56: I think you can just use job_info["output_id"] as it will be None if the job succeeded

src/antares/service/api_services/run_api.py
```diff
@@ -0,0 +1,106 @@
...
+                raise SimulationTimeOutError(job.job_id, time_out)
+            time.sleep(repeat_interval)
+            updated_job = self._get_job_from_id(job.job_id)
+            job.status = updated_job.status
+            job.output_id = updated_job.output_id
+
+        if job.status == JobStatus.FAILED:
+            raise SimulationFailedError(self.study_id)
+
+        if job.unzip_output and job.output_id:
```
MartinBelthle, line 74: I agree here you should assert that we have an output_id. But if we haven't it seems to me we should raise a SimulationFailedError. So i think you should put this as a `or not job.output_id` line 71 and remove it from this line

src/antares/service/api_services/run_api.py
```diff
@@ -0,0 +1,106 @@
...
+            self._wait_unzip_output(self.study_id, ["UNARCHIVE"], job.output_id)
+
+        return None
+
+    def _wait_unzip_output(self, ref_id: str, type: list[str], job_output_id: str) -> None:
```
MartinBelthle, line 79: Actually we shouldn't give ["UNARCHIVE"] to this method, we should use it internally as we cannot call this method with another task_type

tests/antares/services/api_services/test_study_api.py
```diff
@@ -330,3 +335,86 @@ def test_create_link_same_area(self):
...
+                        "status": "pending",
+                        "launcher_params": dumps(parameters.to_api()),
+                    },
+                    "status_code": 200,
+                },
+                {
+                    "json": {
+                        "id": job_id,
+                        "status": "running",
+                        "launcher_params": dumps(parameters.to_api()),
```
MartinBelthle, line 365: You can store `dumps(parameters.to_api())` inside a variable instead of calling it 4 times

tests/antares/services/api_services/test_study_api.py
```diff
@@ -330,3 +335,86 @@ def test_create_link_same_area(self):
...
+            assert job.status == JobStatus.SUCCESS
+
+            response_list.pop()
+            mocker.get(job_url, response_list)
+            mocker.post(tasks_url, status_code=200)
+
+            job = self.study.run_antares_simulation(parameters)
+            with pytest.raises(SimulationTimeOutError):
+                self.study.wait_job_completion(job, time_out=2)
+
```
MartinBelthle, line 404: Perhaps you can separate a bit more these steps with just a line `# =======` + a comment as it's a bit hard to read the whole test

src/antares/service/api_services/run_api.py
```diff
@@ -0,0 +1,106 @@
...
+        job_info = response.json()
+        status = JobStatus.from_str(job_info["status"])
+        output_id = job_info["output_id"] if status == JobStatus.SUCCESS else None
+        launcher_params = json.loads(job_info["launcher_params"])
+        return Job(job_id=job_id, status=status, unzip_output=launcher_params["auto_unzip"], output_id=output_id)
+
+    def wait_job_completion(self, job: Job, time_out: int) -> None:
+        start_time = time.time()
+        repeat_interval = 5
+        while job.status in (JobStatus.RUNNING, JobStatus.PENDING):
```
MartinBelthle, line 63: I think here we should do something about jobs being already a success before entering this method.
As for now we'll reach the line 74 without a job.output_id.
I think we should do a check before the loop with this:
```python            
updated_job = self._get_job_from_id(job.job_id)
job.status = updated_job.status
job.output_id = updated_job.output_id
```
This way we'll have an output id before entering line 74

MartinBelthle: Or we could do it just before line 74 by doing the same code, if not job.output_id: same code

src/antares/service/api_services/run_api.py
```diff
@@ -0,0 +1,106 @@
...
+        try:
+            response = self._wrapper.post(url, json=payload)
+            tasks = response.json()
+            task_id = self._get_task_id(job_output_id, tasks)
+            url += f"/{task_id}"
+            self._get_task_until_success(url, repeat_interval)
+        except APIError as e:
+            raise AntaresSimulationUnzipError(self.study_id, e.message) from e
+
+    def _get_task_id(self, job_output_id: str, tasks: list[dict[str, Any]]) -> str:
```
MartinBelthle, line 92: We could be more specific, like _get_unarchiving_task_id as this code will only work for this type of task and we might one day introduce a get_task_id somewhere in the code that will be more generic

src/antares/service/api_services/run_api.py
```diff
@@ -0,0 +1,106 @@
...
+        for task in tasks:
+            task_name = task["name"]
+            output_id = task_name.split("/")[-1].split(" ")[0]
+            if output_id == job_output_id:
+                return task["id"]
+        raise AntaresSimulationUnzipError(self.study_id, "Could not find task for unarchiving job")
+
+    def _get_task_until_success(self, url: str, repeat_interval: int) -> None:
+        task_success = False
```
MartinBelthle, line 101: This task can fail. And if it does you will never go out of this while. You can introduce a more generic method that has a task_id as an argument and builds its url internally. it could be called `wait_task_completion` and will have the same behavior as wait_job_completion: a timeout and raise a Timeout error etc.. (For now you can make it private but we'll need this method at some point). 

Also task["result"] can return None so you'll have to handle this as a sign it's still running to avoid KeyErrors

tests/antares/services/api_services/test_study_api.py
```diff
@@ -330,3 +335,86 @@ def test_create_link_same_area(self):
             match=f"Could not create the link {area} / {area}: A link cannot start and end at the same area",
         ):
             self.study.create_link(area_from=area, area_to=area)
+
+    def test_run_and_wait_antares_simulation(self):
+        parameters = AntaresSimulationParameters(solver=Solver.COIN, nb_cpu=2, unzip_output=True, presolve=False)
+
+        # patch simulates the repeating intervals so that we don't have to wait X seconds during the tests
```
MartinBelthle, line 342: Smart

MartinBelthle (changes requested):
Some comments but we're in the right direction :)

As we discussed you can introduce a test inside test_web_client that just launches a simulation and asserts the job succeeded

src/antares/service/api_services/run_api.py
salemsd: `job_info["output_id"]` doesn't even exist before the job succeeds, so it will throw a KeyError

src/antares/model/simulation.py
```diff
@@ -0,0 +1,65 @@
...
+    SUCCESS = "success"
+    FAILED = "failed"
+
+    @staticmethod
+    def from_str(input: str) -> "JobStatus":
+        return JobStatus.__getitem__(input.upper())
+
+
+class Job(BaseModel):
+    job_id: str
```
MartinBelthle, line 62: I just realized we should remove the `unzip_output` from this class and instead add an argument `parameters` that would be an `AntaresSimulationParameters`. This way we can still access the value and the user will have more info on his job. Because with the current way we do things, after launching a job the user cannot find the options he launched it with

tests/integration/test_web_client.py
```diff
@@ -497,3 +498,16 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop):
...
+
+        job = study.run_antares_simulation(parameters)
+        assert isinstance(job, Job)
+        assert job.status == JobStatus.PENDING
+
+        study.wait_job_completion(job, time_out=60)
+        assert job.status == JobStatus.SUCCESS
+
+        assert job.output_id is not None
+        assert job.unzip_output is True
```
MartinBelthle, line 513: Add test for the new attribute `parameter`

src/antares/service/api_services/run_api.py
```diff
@@ -0,0 +1,120 @@
...
+        except APIError as e:
+            raise AntaresSimulationUnzipError(self.study_id, e.message) from e
+
+    def _get_unarchiving_task_id(self, job_output_id: str, tasks: list[dict[str, Any]]) -> str:
+        for task in tasks:
+            task_name = task["name"]
+            output_id = task_name.split("/")[-1].split(" ")[0]
+            if output_id == job_output_id:
+                return task["id"]
+        raise AntaresSimulationUnzipError(self.study_id, "Could not find task for unarchiving job")
```
MartinBelthle, line 104: We should add the job id at the end of the log. It could help the user and also us debugging

MartinBelthle (changes requested):
Minor changes, mainly a spec change I just came up with :/

src/antares/service/api_services/run_api.py
MartinBelthle: Okay. Could be `output_id = job_info.get("output_id")` then ?

src/antares/service/api_services/run_api.py
```diff
@@ -0,0 +1,125 @@
...
+            self._update_job(job)
+
+        while job.status in (JobStatus.RUNNING, JobStatus.PENDING):
+            if time.time() - start_time > time_out:
+                raise SimulationTimeOutError(job.job_id, time_out)
+            time.sleep(repeat_interval)
+            self._update_job(job)
+
+        if job.status == JobStatus.FAILED or not job.output_id:
+            raise SimulationFailedError(self.study_id)
```
MartinBelthle, line 74: I think we should also log the job_id inside this Exception for the user and debugging purpose

src/antares/service/api_services/run_api.py
```diff
@@ -0,0 +1,125 @@
...
+    def _wait_unzip_output(self, ref_id: str, job: Job, time_out: int) -> None:
+        url = f"{self._base_url}/tasks"
+        repeat_interval = 2
+        payload = {"type": ["UNARCHIVE"], "ref_id": ref_id}
+        try:
+            response = self._wrapper.post(url, json=payload)
+            tasks = response.json()
+            task_id = self._get_unarchiving_task_id(job, tasks)
+            self._wait_task_completion(task_id, repeat_interval, time_out)
+        except APIError as e:
```
MartinBelthle, line 98: I think you can simply do a
```python
except Exception as e:
    raise AntaresSimulationUnzipError(self.study_id, job.job_id, e.message) from e
```
as I think we want to raise this issue in both cases

MartinBelthle (changes requested):
Almost done :)

MartinBelthle (approved):

PR #36, opened 2024-12-13
Fix key_error issue

killian-scalian:
- Fix key_error issue
- remove area_id in some useless places


MartinBelthle (approved):
Thanks for the fix. I'll let you write the test you like and if theres' no other file involved you can merge

PR #37, opened 2024-12-17
feat(api): add outputs

salemsd:
- Added outputs and service to study
- Added unit and integration tests


tests/antares/services/api_services/test_study_api.py
```diff
@@ -52,6 +53,184 @@ class TestCreateAPI:
         ServiceFactory(api, study_id).create_thermal_service(),
         ServiceFactory(api, study_id).create_renewable_service(),
     )
+    json_output = [
```
salemsd, line 56: Wasn't sure about the json for the outputs in the unit tests. Since it's very large, should I just include the relevant keys or leave it like that

src/antares/model/study.py
```diff
@@ -367,6 +371,35 @@ def wait_job_completion(self, job: Job, time_out: int = 172800) -> None:
...
+        self._outputs = {output.name: output for output in outputs}
+        return outputs
+
+    def get_outputs(self) -> dict[str, Output]:
+        """
+        Get outputs of current study
+
+        Returns: (output_id, Output) dictionary
+        """
+        return self._outputs.copy()
```
MartinBelthle, line 390: Instead of doing a copy what we do is we return a MappingProxyType. This basically does the same thing but we should keep the code coherent (see `get_binding_constraints` method for instance)

src/antares/model/study.py
```diff
@@ -367,6 +371,35 @@ def wait_job_completion(self, job: Job, time_out: int = 172800) -> None:
...
+    def get_output(self, output_id: str) -> Optional[Output]:
+        """
+        Get a specific output
+
+        Args:
+            output_id: id of the output to get
+
+        Returns: Output with the output_id
+        """
+        return self._outputs.get(output_id)
```
MartinBelthle, line 401: Should be crash instead of returning a None ? I don't know

src/antares/service/api_services/output_api.py
```diff
@@ -0,0 +1,37 @@
...
+        self.study_id = study_id
+        self._base_url = f"{self.config.get_host()}/api/v1"
+        self._wrapper = RequestWrapper(self.config.set_up_api_conf())
+
+    def read_outputs(self) -> list[Output]:
+        url = f"{self._base_url}/studies/{self.study_id}/outputs"
+        try:
+            response = self._wrapper.get(url)
+            outputs_json_list = response.json()
+            outputs_list: list[Output] = list()
```
MartinBelthle, line 32: Could be a comprehension list: 
```python
outputs_list = [Output(name=output["name"], archived=output["archived"]) for output in outputs_json_list]
```

tests/integration/test_web_client.py
```diff
@@ -499,7 +499,12 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop):
             new_study.get_binding_constraints().keys()
         )
 
-        # ===== test run study simulation and wait job completion ======
+        # ===== Test outputs (Part 1) =====
```
MartinBelthle, line 502: Add - before simulation

tests/integration/test_web_client.py
```diff
@@ -512,3 +517,11 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop):
         assert job.output_id is not None
         assert job.parameters == parameters
         assert job.parameters.unzip_output is True
+
+        # ===== Test outputs (Part 2 - after simulation) =====
+
+        study.read_outputs()
+        outputs = study.get_outputs()
+        assert len(outputs) > 0
```
MartinBelthle, line 525: You can check that you have only one and that it's unarchived

tests/antares/services/api_services/test_study_api.py
MartinBelthle: Indeed we don't need all this. You can just simplify by:
``` python
{
    "name": "20241217-1115eco-sdqsd",
    "type": "economy",
    "settings": {},
    "archived": False,
}
```
for instance

MartinBelthle (changes requested):
Minor changes, otherwise it's good :)

src/antares/model/study.py
```diff
@@ -367,6 +371,37 @@ def wait_job_completion(self, job: Job, time_out: int = 172800) -> None:
...
+        Get a specific output
+
+        Args:
+            output_id: id of the output to get
+
+        Returns: Output with the output_id
+
+        Raises: KeyError if it doesn't exist
+        """
+        return self._outputs[f"{output_id}"]
```
MartinBelthle, line 403: could just be self._outputs[output_id]

src/antares/service/api_services/output_api.py
```diff
@@ -0,0 +1,38 @@
...
+        self.study_id = study_id
+        self._base_url = f"{self.config.get_host()}/api/v1"
+        self._wrapper = RequestWrapper(self.config.set_up_api_conf())
+
+    def read_outputs(self) -> list[Output]:
+        url = f"{self._base_url}/studies/{self.study_id}/outputs"
+        try:
+            response = self._wrapper.get(url)
+            outputs_json_list = response.json()
+            if isinstance(outputs_json_list, dict):
```
MartinBelthle, line 34: the endpoint is supposed to return a list so I don't see where you could enter this if

tests/antares/services/api_services/test_study_api.py
```diff
@@ -215,6 +217,15 @@ def test_read_study_api(self):
             mocker.get(renewable_url, json=[])
             mocker.get(thermal_url, json=[])
             mocker.get(storage_url, json=[])
+            mocker.get(
+                output_url,
```
MartinBelthle, line 221: For these 2 tests, you can just return an empty list: `mocker.get(output_url, json=[])` as you don't really test the content

tests/antares/services/api_services/test_study_api.py
```diff
@@ -425,3 +448,39 @@ def test_run_and_wait_antares_simulation(self):
...
+                },
+            ]
+            mocker.get(run_url, json=json_output, status_code=200)
+
+            self.study.read_outputs()
+
+            assert len(self.study.get_outputs()) == 2
+            output1 = self.study.get_output(json_output[0].get("name"))
+            output2 = self.study.get_output(json_output[1].get("name"))
+            assert output1 is not None
```
MartinBelthle, line 477: I think now you can return the checks of nullity

tests/integration/test_web_client.py
```diff
@@ -512,3 +517,12 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop):
         assert job.output_id is not None
         assert job.parameters == parameters
         assert job.parameters.unzip_output is True
+
+        # ===== Test outputs (Part 2 - after simulation) =====
+
+        output = study.read_outputs()[0]
+        outputs = study.get_outputs()
+        assert len(outputs) == 1
+        assert outputs.get(output.name).archived is not False
```
MartinBelthle, line 526: Could be simplified in ` assert outputs.get(output.name).archived` I think

MartinBelthle (changes requested):

src/antares/service/api_services/output_api.py
salemsd: My bad i misread the api doc

MartinBelthle (approved):

PR #42, opened 2024-12-19
feat(api): add output funcionalities (get_matrix, aggregate_values)

salemsd:
- Added more functionalities for outputs
- Added unit and integration tests


src/antares/craft/exceptions/exceptions.py
```diff
@@ -12,6 +12,8 @@
 
 from typing import List
 
+from antares.craft.model.output import McType, ObjectType
```
MartinBelthle, line 15: Generally speaking I think we prefer not having to import anything related to the project in this file as it's like an util file

src/antares/craft/model/output.py
```diff
@@ -9,9 +9,96 @@
...
 # This file is part of the Antares project.
+from enum import Enum
+from typing import Any
+
+import pandas as pd
+
 from pydantic import BaseModel
 
 
+class QueryFile(Enum):
```
MartinBelthle, line 20: You're missing the value ID = "id"
I would also prefer to rename STSTORAGE ST_STORAGE

src/antares/craft/model/output.py
```diff
@@ -9,9 +9,96 @@
...
+        columns_names: names or regexes (if query_file is of type details) to select columns (comma separated)
+    """
+
+    query_file: QueryFile
+    frequency: Frequency
+    mc_years: str = ""
+    type_ids: str = ""
+    columns_names: str = ""
+
+    def to_query(self, object_type: ObjectType) -> str:
```
MartinBelthle, line 63: Rename to_api_query to show it's for api

src/antares/craft/model/output.py
```diff
@@ -9,9 +9,96 @@
...
+        query_file: The file to query.
+        frequency: "hourly", "daily", "weekly", "monthly", "annual"
+        mc_years: Monte Carlo years to include in the query. If left empty, all years are included.
+        type_ids: which links/areas to be selected (ex: "be - fr"). If empty, all are selected (comma separated)
+        columns_names: names or regexes (if query_file is of type details) to select columns (comma separated)
+    """
+
+    query_file: QueryFile
+    frequency: Frequency
+    mc_years: str = ""
```
MartinBelthle, line 59: Probably best to type it as a list for users. Same for column_names

src/antares/craft/model/output.py
```diff
@@ -9,9 +9,96 @@
...
+        columns_names = f"&columns_names={self.columns_names}" if self.columns_names else ""
+
+        return f"query_file={self.query_file.value}&frequency={self.frequency.value}{mc_years}{type_ids}{columns_names}&format=csv"
+
+
 class Output(BaseModel):
     name: str
     archived: bool
+
+    def __init__(self, output_service, **kwargs: Any):  # type: ignore
```
MartinBelthle, line 75: I think you should remove the usage of pydantic inside this class and just do a normal class

src/antares/craft/model/output.py
```diff
@@ -9,9 +9,96 @@
...
+
+        Args:
+            path: output path, eg: "mc-all/areas/south/values-hourly"
+
+        Returns: Pandas DataFrame
+        """
+        full_path = f"output/{self.name}/economy/{path}"
+        return self._output_service.get_matrix(full_path)
+
+    def aggregate_values(
```
MartinBelthle, line 91: Now that I see the code I don't know if we should keep one method with arguments or 4 distincts method. The benefit of having 4 method would be to not introduce McType and ObjectType class that I don't really like and also ensure the queryfiles can always work. I'll see with Alexander what he thinks about it

src/antares/craft/model/study.py
```diff
@@ -219,8 +219,8 @@ def __init__(
         self._area_service = service_factory.create_area_service()
         self._link_service = service_factory.create_link_service()
         self._run_service = service_factory.create_run_service()
-        self._output_service = service_factory.create_output_service()
         self._binding_constraints_service = service_factory.create_binding_constraints_service()
+        self._output_service = service_factory.create_output_service()
```
MartinBelthle, line 223: Don't think you need to create an output service here anymore

src/antares/craft/service/api_services/output_api.py
```diff
@@ -26,11 +29,15 @@ def __init__(self, config: APIconf, study_id: str):
...
+
+    def aggregate_values(
+        self, output_id: str, aggregation_entry: AggregationEntry, mc_type: McType, object_type: ObjectType
+    ) -> pd.DataFrame:
+        url = f"{self._base_url}/studies/{self.study_id}/{object_type.value}/aggregate/{mc_type.value}/{output_id}?{aggregation_entry.to_query(object_type)}"
         try:
             response = self._wrapper.get(url)
-            outputs_json_list = response.json()
-            return [Output(name=output["name"], archived=output["archived"]) for output in outputs_json_list]
+            return pd.read_csv(StringIO(response.text))
```
MartinBelthle, line 41: In AntaresWeb we do `pd.read_csv(io.BytesIO(response.content), sep=",")`, don't know if it's faster. We can keep the code as is but don't you have issue as you didn't specify the separator ?

tests/antares/services/api_services/test_study_api.py
```diff
@@ -472,3 +474,36 @@ def test_read_outputs(self):
             mocker.get(run_url, json={"description": error_message}, status_code=404)
             with pytest.raises(OutputsRetrievalError, match=error_message):
                 self.study.read_outputs()
+
+            # ==== get_matrix() ====
```
MartinBelthle, line 478: You should probably put this in a separated test as you're not really checking the same thing as the first part of the test and you can re-introduce a fake output at the start of your test

tests/antares/services/api_services/test_study_api.py
```diff
@@ -472,3 +474,36 @@ def test_read_outputs(self):
...
+
+            matrix_url = f"https://antares.com/api/v1/studies/{self.study_id}/raw?path=output/{output1.name}/economy/mc-all/grid/links"
+            matrix_output = {"columns": ["upstream", "downstream"], "data": [["be", "fr"]]}
+            mocker.get(matrix_url, json=matrix_output)
+
+            matrix = output1.get_matrix("mc-all/grid/links")
+            assert isinstance(matrix, pd.DataFrame)
+            assert matrix.shape[1] == len(matrix_output["columns"])
```
MartinBelthle, line 486: Can't you just put matrix_output inside a dataframe and test the equality ? It would be clearer
Same remark for the aggregate values part

tests/integration/test_web_client.py
```diff
@@ -526,3 +527,17 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop):
         assert not outputs.get(output.name).archived
         study_with_outputs = read_study_api(api_config, study._study_service.study_id)
         assert study_with_outputs.get_outputs() == outputs
+
+        # ===== Output get_matrix =====
+
+        matrix = output.get_matrix("mc-all/grid/links")
+
+        assert isinstance(matrix, pd.DataFrame)
+        assert not matrix.empty
```
MartinBelthle, line 536: Could you check the value a bit more ?
Same for the aggregate values test

MartinBelthle (changes requested):

src/antares/craft/service/api_services/output_api.py
salemsd: According to the doc, separator is "," by default, so no problem. As for the data handling, apparently they offer similar performance -> `StringIO, however, is a native in-memory unicode container and will exhibit similar speed to BytesIO`, from python doc

MartinBelthle: Okay. It's weird as in pandas doc online I see , as default but in the source code it seems like it's None. Weird

src/antares/craft/model/output.py
MartinBelthle: Alexander doesn't care. So I think it's better to introduce 4 methods. You'll also have to introduce other query classes that go for every method, see the AntaresWeb code below: 
```python
class MCIndAreasQueryFile(StrEnum):
    VALUES = "values"
    DETAILS = "details"
    DETAILS_ST_STORAGE = "details-STstorage"
    DETAILS_RES = "details-res"


class MCAllAreasQueryFile(StrEnum):
    VALUES = "values"
    DETAILS = "details"
    DETAILS_ST_STORAGE = "details-STstorage"
    DETAILS_RES = "details-res"
    ID = "id"


class MCIndLinksQueryFile(StrEnum):
    VALUES = "values"


class MCAllLinksQueryFile(StrEnum):
    VALUES = "values"
    ID = "id"

```

MartinBelthle: I think the wording FileType would be more appropriate as we'll also use it for the local implem one day

MartinBelthle: From an user perspective these classes will probably be hard to use. We'll have to think if there's a better solution in a next PR

src/antares/craft/model/study.py
salemsd: The alternative would be to create the output service in local and api manually without the factory. Same thing could be said about binding_constraint_service which is not really needed in study either.
Or we could move read_outputs back into output_service but that would create redundancy with an output also having its own service.
I don't see what else I could do ?

MartinBelthle: I think what you need to do is create the StudyService by giving it an output_service and with that you won't have to give the service inside the method

MartinBelthle: I'll check

MartinBelthle: I think what you should do is add a method set_output_service inside the study service that you'll call inside the create_study_service method. This way you can remove the service from the study as you'll have it inside the study_service. That's something we're doing already for the area_service if you want to check

src/antares/craft/service/base_services.py
```diff
@@ -631,10 +644,29 @@ def wait_job_completion(self, job: Job, time_out: int) -> None:
...
+        Creates a matrix of aggregated raw data
+
+        Args:
+            output_id: id of the output
+            aggregation_entry: input for the /aggregate endpoint
+            mc_type: all or ind (enum)
+            object_type: links or areas (enum)
+
+        Returns: Pandas DataFrame corresponding to the aggregated raw data
+        """
```
MartinBelthle, line 672: You forgot the `pass`. Don't know if it's needed but we always add it

src/antares/craft/service/base_services.py
```diff
@@ -631,10 +644,29 @@ def wait_job_completion(self, job: Job, time_out: int) -> None:
...
+    @abstractmethod
+    def aggregate_values(
+        self, output_id: str, aggregation_entry: AggregationEntry, object_type: str, mc_type: str
+    ) -> pd.DataFrame:
+        """
+        Creates a matrix of aggregated raw data
+
+        Args:
+            output_id: id of the output
+            aggregation_entry: input for the /aggregate endpoint
```
MartinBelthle, line 667: We shouldn't talk about endpoint inside the base service as it refers to the api part

tests/integration/test_web_client.py
MartinBelthle: Unresolved

src/antares/craft/model/output.py
```diff
@@ -9,9 +9,142 @@
...
+
+        Args:
+            path: output path, eg: "mc-all/areas/south/values-hourly"
+
+        Returns: Pandas DataFrame
+        """
+        full_path = f"output/{self.name}/economy/{path}"
+        return self._output_service.get_matrix(full_path)
+
+    def aggregate_values_areas_mc_ind(self, aggregation_entry: AggregationEntry) -> pd.DataFrame:
```
MartinBelthle, line 108: Nice to introduce such methods but I still have a few remarks.
- I think you can remove the `_values` from each method name
- I don't really like the way an user has to call these methods. 
Currently it's 
```python
aggregation_entry = AggregationEntry(query_file=MCIndAreas.VALUES, frequency=Frequency.ANNUAL)
aggregated_matrix = output.aggregate_values_areas_mc_ind(aggregation_entry)
```
I would prefer to do: 
```python
aggregated_matrix = output.aggregate_areas_mc_ind(query_file="values", frequency="annual")
```
and then on our side inside the method ensure the frequency and the query_file are okay (via the usage of an internal object)
This will avoid users issue (example, giving MC_years when doing mc_all aggregation which doesn't make sense)

src/antares/craft/model/output.py
```diff
@@ -9,9 +9,142 @@
...
+        """
+        Gets the matrix of the output
+
+        Args:
+            path: output path, eg: "mc-all/areas/south/values-hourly"
+
+        Returns: Pandas DataFrame
+        """
+        full_path = f"output/{self.name}/economy/{path}"
+        return self._output_service.get_matrix(full_path)
```
MartinBelthle, line 106: Could be better to hide this inside the implem and just give the output_id and the file_path to the method

src/antares/craft/model/output.py
```diff
@@ -9,9 +9,142 @@
...
+        return f"query_file={self.query_file.value}&frequency={self.frequency.value}{mc_years}{type_ids}{columns_names}&format=csv"
+
+
+class Output:
+    def __init__(self, name: str, archived: bool, output_service):  # type: ignore
+        self._name = name
+        self._archived = archived
+        self._output_service = output_service
+
+    def __eq__(self, other: Any) -> bool:
```
MartinBelthle, line 83: Don't really see the point, you can remove it

MartinBelthle (changes requested):

src/antares/craft/model/output.py
salemsd: When comparing two output lists (ex: in the integration tests), each output is compared by its memory address instead of its attributes. That's why I overrode the __eq__ method

tests/integration/test_web_client.py
```diff
@@ -526,3 +526,26 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop):
...
+        assert matrix.shape == (1, 2)
+        assert "upstream" in matrix
```
MartinBelthle, line 537: Couldn't just you check the whole matrix as it's a bit odd to understand the test

src/antares/craft/model/output.py
MartinBelthle: Not really important but I don't think someone will ever use this. Besides we could have issues with this. If 2 studies have 2 different outputs with the same name, the 2 outputs are in fact not equal.

MartinBelthle (changes requested):
Really minor comments, almost done

MartinBelthle (approved):

PR #44, opened 2025-01-06
Feat/read links

mehdiwahada:


tests/antares/services/api_services/test_link_api.py
```diff
@@ -29,6 +29,44 @@
...
 
 
+@pytest.fixture
+def expected_link():
+    area_from_name = "zone1 auto"
+    area_to_name = "zone4auto"
+    api = APIconf("https://antares.com", "token", verify=False)
+    study_id = "22c52f44-4c2a-407b-862b-490887f93dd8"
+    link_service = ServiceFactory(api, study_id).create_link_service()
+    properties = {
```
MartinBelthle, line 39: You could add this inside the class instead of having to create a fixture as we already have a link defined there

src/antares/craft/service/api_services/link_api.py
```diff
@@ -214,7 +214,46 @@ def create_capacity_indirect(self, series: pd.DataFrame, area_from: str, area_to
             raise LinkUploadError(area_from, area_to, "indirectcapacity", e.message) from e
 
     def read_links(self) -> list[Link]:
-        raise NotImplementedError
+        url = f"{self._base_url}/studies/{self.study_id}/links"
+        json_links = self._wrapper.get(url).json()
+        links = []
+
+        for link in json_links:
+            link_area_from_id = link.pop("area1")
```
MartinBelthle, line 222: This works well but we could probably put this inside a utility method as we may have to re use it

MartinBelthle (changes requested):

tests/antares/services/api_services/test_link_api.py
```diff
@@ -267,3 +299,35 @@ def test_get_direct_capacity_fail(self):
                 match=f"Could not download directcapacity matrix for link {self.area_from.id}/{self.area_to.id}",
             ):
                 self.link.get_capacity_direct()
+
+    def test_read_links(self, expected_link):
```
MartinBelthle, line 303: You could add a test that fails to ensure we raise the right exception with the right message (see test_get_direct_capacity_fail` that you wrote for instance)

tests/antares/services/api_services/test_link_api.py
```diff
@@ -267,3 +299,35 @@ def test_get_direct_capacity_fail(self):
...
+                "area1": "zone1 auto",
+                "area2": "zone4auto",
+            }
+        ]
+
+        with requests_mock.Mocker() as mocker:
+            expected_link_list = [expected_link]
+            mocker.get(url_read_links, json=json_links)
+            actual_link_list = self.study.read_links()
+            assert len(actual_link_list) == len(expected_link_list)
```
MartinBelthle, line 330: You could assert that the list contains one link and after declare a variable instead of having `actual_link_list[0]` in your code multiple times. Also i think you can remove `expected_link_list` and just use expected_link as is

tests/integration/test_web_client.py
```diff
@@ -392,6 +392,14 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop):
         thermal_fr.update_properties(new_props)
         assert thermal_fr.properties.group == ThermalClusterGroup.NUCLEAR
 
+        # assert study got all links
+        links = study.read_links()
```
MartinBelthle, line 396: Could you also add a little test on the properties for one link ?
And also really minor but I think it's more readable to put `assert len(links) == 2` before `test_link_be_fr = links[0]` than after

MartinBelthle (changes requested):
Don't forget to rebase with main :)

src/antares/craft/service/api_services/link_api.py
```diff
@@ -214,7 +215,52 @@ def create_capacity_indirect(self, series: pd.DataFrame, area_from: str, area_to
...
+            "assetType": "asset-type",
+        }
+
+        link = {mapping.get(k, k): v for k, v in link.items()}
+        link["filter-synthesis"] = set(link["filter-synthesis"].split(", "))
+        link["filter-year-by-year"] = set(link["filter-year-by-year"].split(", "))
+        link_properties = LinkProperties(**link)
+        return Link(link_area_from_id, link_area_to_id, self, link_properties, link_ui)
+
+    def read_study_links(self, link_list: list) -> list[Link]:
```
MartinBelthle, line 257: I think we can remove this method and just put the for loop inside the `read_links` method

src/antares/craft/service/api_services/link_api.py
```diff
@@ -214,7 +215,52 @@ def create_capacity_indirect(self, series: pd.DataFrame, area_from: str, area_to
...
+            url = f"{self._base_url}/studies/{self.study_id}/links"
+            json_links = self._wrapper.get(url).json()
+            links = self.read_study_links(json_links)
+
+            links.sort(key=lambda link_obj: link_obj.area_from_id)
+        except APIError as e:
+            raise LinksRetrievalError(self.study_id, e.message) from e
+        return links
+
+    def read_link(self, link: dict[str, Any]) -> Link:
```
MartinBelthle, line 228: We could rename this method to make it more clear. Something like `convert_api_link_to_internal_link`. And also rename the method parameter `link` in `api_link` to make it even clearer

MartinBelthle (changes requested):
Almost done :)

MartinBelthle (approved):

PR #53, opened 2025-01-13
feat(api): add possibility to move a study

salemsd:
- Added move(folder_path) method to move a study to a new parent folder

### DIFF
--- a/src/antares/craft/exceptions/exceptions.py
+++ b/src/antares/craft/exceptions/exceptions.py
@@ -250,6 +250,12 @@ class StudyVariantCreationError(Exception):
         super().__init__(self.message)
 
 
+class StudyMoveError(Exception):
+    def __init__(self, new_folder_name: str, message: str) -> None:
+        self.message = f"Could not move the study to folder {new_folder_name}: " + message
+        super().__init__(self.message)
+
+
 class ThermalMatrixDownloadError(Exception):
     def __init__(self, area_name: str, cluster_name: str, matrix_name: str, message: str) -> None:
         self.message = (
--- a/src/antares/craft/exceptions/exceptions.py
+++ b/src/antares/craft/exceptions/exceptions.py
@@ -251,8 +251,8 @@ class StudyVariantCreationError(Exception):
 
 
 class StudyMoveError(Exception):
-    def __init__(self, new_folder_name: str, message: str) -> None:
-        self.message = f"Could not move the study to folder {new_folder_name}: " + message
+    def __init__(self, study_id: str, new_folder_name: str, message: str) -> None:
+        self.message = f"Could not move the study {study_id} to folder {new_folder_name}: " + message
         super().__init__(self.message)
 
 
--- a/src/antares/craft/model/study.py
+++ b/src/antares/craft/model/study.py
@@ -82,7 +82,7 @@ def create_study_api(
             study.move(parent_path)
             url = f"{base_url}/studies/{study_id}"
             json_study = wrapper.get(url).json()
-            study.path = json_study.pop("folder")
+            study.path = PurePath(json_study.pop("folder"))
         return study
     except (APIError, StudyMoveError) as e:
         raise StudyCreationError(study_name, e.message) from e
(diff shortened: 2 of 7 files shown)


src/antares/craft/exceptions/exceptions.py
```diff
@@ -250,6 +250,12 @@ def __init__(self, study_name: str, message: str) -> None:
         super().__init__(self.message)
 
 
+class StudyMoveError(Exception):
+    def __init__(self, new_folder_name: str, message: str) -> None:
+        self.message = f"Could not move the study to folder {new_folder_name}: " + message
```
MartinBelthle, line 255: Add the study id inside the log

src/antares/craft/model/study.py
```diff
@@ -73,10 +73,13 @@ def create_study_api(
         response = wrapper.post(url)
         study_id = response.json()
         study_settings = _returns_study_settings(base_url, study_id, wrapper, False, settings)
+        url = f"{base_url}/studies/{study_id}"
```
MartinBelthle, line 76: We should do something here:
- Either we don't allow the user to create inside a specific repo and we don't need this, the folder will always be None
- Or we allow the user to specify a folder (I prefer this option) and we keep the code as it is (the attribute `folder` is always inside the json_study)

src/antares/craft/model/study.py
```diff
@@ -143,6 +146,7 @@ def create_study_local(
         version=version,
         service_factory=ServiceFactory(config=local_config, study_name=study_name),
         settings=local_settings,
+        path=parent_directory,
```
MartinBelthle, line 149: Should be study_directory instead. Same for line 180

src/antares/craft/model/study.py
```diff
@@ -184,9 +189,10 @@ def read_study_api(api_config: APIconf, study_id: str) -> "Study":
 
     study_name = json_study.pop("name")
     study_version = str(json_study.pop("version"))
+    path = json_study.pop("folder") if "folder" in json_study else None
```
MartinBelthle, line 192: Remove the if else

src/antares/craft/model/study.py
```diff
@@ -217,9 +223,11 @@ def __init__(
         version: str,
         service_factory: ServiceFactory,
         settings: Union[StudySettings, StudySettingsLocal, None] = None,
+        path: Union[str, PurePath, None] = None,
```
MartinBelthle, line 226: Should just be a PurePath. All the cast etc. should be done inside the implem

src/antares/craft/model/study.py
```diff
@@ -421,6 +429,10 @@ def delete_output(self, output_name: str) -> None:
         self._study_service.delete_output(output_name)
         self._outputs.pop(output_name)
 
+    def move(self, destination_folder: Path) -> None:
+        self._study_service.move_study(destination_folder)
```
MartinBelthle, line 433: We should be more clear here to specify the destination_folder corresponds to the parent path in the end (rename the arg).
Also, the next line should be done inside the implems, the method should return directly the study_path 
```
self.path = self._study_service.move_study(destination_folder)
```

tests/antares/services/api_services/test_study_api.py
```diff
@@ -74,11 +76,36 @@ def test_create_study_test_ok(self) -> None:
             mocker.post(expected_url, json=self.study_id, status_code=200)
             config_urls = re.compile(f"https://antares.com/api/v1/studies/{self.study_id}/config/.*")
             mocker.get(config_urls, json={}, status_code=200)
+            expected_url_path = f"https://antares.com/api/v1/studies/{self.study_id}"
+            mocker.get(
+                expected_url_path,
+                json={
```
MartinBelthle, line 82: Can't you simplify this json to just keep important keys ?

MartinBelthle (changes requested):

src/antares/craft/model/study.py
```diff
@@ -73,10 +77,15 @@ def create_study_api(
...
         study_id = response.json()
         study_settings = _returns_study_settings(base_url, study_id, wrapper, False, settings)
-
-    except APIError as e:
+        study = Study(study_name, version, ServiceFactory(api_config, study_id), study_settings)
+        if parent_path:
+            study.move(parent_path)
+            url = f"{base_url}/studies/{study_id}"
+            json_study = wrapper.get(url).json()
+            study.path = PurePath(json_study.pop("folder"))
```
MartinBelthle, line 85: For safety, I think you should add the same `if` as in line 199 as we could have a folder which is at None

src/antares/craft/service/api_services/study_api.py
```diff
@@ -167,6 +169,15 @@ def delete_output(self, output_name: str) -> None:
         except APIError as e:
             raise OutputDeletionError(self.study_id, output_name, e.message) from e
 
+    def move_study(self, new_parent_path: Path) -> PurePath:
+        url = f"{self._base_url}/studies/{self.study_id}/move?folder_dest={new_parent_path}"
+        try:
+            self._wrapper.put(url)
+            json_study = self._wrapper.get(f"{self._base_url}/studies/{self.study_id}").json()
+            return PurePath(json_study.pop("folder"))
```
MartinBelthle, line 177: Same remark about the None

MartinBelthle (approved):

PR #67, opened 2025-01-28
fix(local): get thermal matrix checked the wrong path

killian-scalian:
The name of the folder under area_id is properties.group.value, seems it is thermal.name instead

### DIFF
--- a/src/antares/craft/service/local_services/thermal_local.py
+++ b/src/antares/craft/service/local_services/thermal_local.py
@@ -54,7 +54,7 @@ class ThermalLocalService(BaseThermalService):
             time_serie_type,
             self.config.study_path,
             area_id=thermal_cluster.area_id,
-            cluster_id=thermal_cluster.properties.group.value,
+            cluster_id=thermal_cluster.name,
         )
 
     def read_thermal_clusters(self, area_id: str) -> List[ThermalCluster]:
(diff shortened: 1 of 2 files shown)


src/antares/craft/service/local_services/thermal_local.py
```diff
@@ -54,7 +54,7 @@ def get_thermal_matrix(self, thermal_cluster: ThermalCluster, ts_name: ThermalCl
             time_serie_type,
             self.config.study_path,
             area_id=thermal_cluster.area_id,
-            cluster_id=thermal_cluster.properties.group.value,
+            cluster_id=thermal_cluster.name,
```
MartinBelthle, line 57: Should be .id

tests/antares/services/local_services/test_area.py
```diff
@@ -927,9 +927,7 @@ def test_read_thermals_local(self, local_study_w_thermal):
                 assert thermal.properties.cost_generation.value == "SetManually"
 
                 # Create folder and file for timeserie.
-                cluster_path = (
-                    study_path / "input" / "thermal" / "series" / Path(area.id) / Path(thermal.properties.group.value)
-                )
+                cluster_path = study_path / "input" / "thermal" / "series" / Path(area.id) / Path(thermal.name)
```
MartinBelthle, line 930: Should be .id
Same for the other test you changed

MartinBelthle (changes requested):
Indeed it didn't make sense

MartinBelthle (approved):

PR #72, opened 2025-01-29
feat(api): added import_study method to study_api.py

mehdiwahada:


MartinBelthle (changes requested):
Few remarks:
- The destination_path should be optional if the person just want to import the study. If it's None, we should'nt call the move method
- The `import_study` method should be renamed `import_study_api` and put at the same place as `create_study_api`as you won't do study.import_study(...). This also means the method expects a study_path, and an api_config parameter

tests/integration/test_web_client.py
```diff
@@ -92,6 +94,26 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop):
         area_fr.create_wind(ts_matrix)
         assert area_fr.get_wind_matrix().equals(ts_matrix)
 
+        # testing import study
+        test_path = Path(antares_web.desktop_path.joinpath("internal_studies").joinpath(study.service.study_id))
+        copy_dir = tmp_path / test_path.name
+
+        tmp_path_zip = tmp_path / f"{copy_dir.name}"
```
MartinBelthle, line 101: You can remove the f-string:
`tmp_path_zip = tmp_path / copy_dir.name`

tests/integration/test_web_client.py
```diff
@@ -92,6 +94,26 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop):
         area_fr.create_wind(ts_matrix)
         assert area_fr.get_wind_matrix().equals(ts_matrix)
 
+        # testing import study
```
MartinBelthle, line 97: Now that this test work we should put everthing you wrote here at the end of the test as it would make the test clearer

tests/integration/test_web_client.py
```diff
@@ -92,6 +94,26 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop):
...
+        test_path = Path(antares_web.desktop_path.joinpath("internal_studies").joinpath(study.service.study_id))
```
MartinBelthle, line 98: Add a little comment saying why we're doing this weird thing

src/antares/craft/model/study.py
```diff
@@ -89,6 +95,32 @@ def create_study_api(
...
+def import_study_api(api_config: APIconf, study_path: Path, destination_path: Optional[Path]) -> "Study":
+    def has_valid_extension(path: Path) -> bool:
+        valid_extensions = {".zip", ".7z"}
+        return path.suffix in valid_extensions
+
+    session = api_config.set_up_api_conf()
+    wrapper = RequestWrapper(session)
+    base_url = f"{api_config.get_host()}/api/v1"
+
+    if has_valid_extension(study_path):
```
MartinBelthle, line 107: I think it will be clearer to first do a 
`if not has_valid_extension: raise` and then continue what you did without the if.
Also the exception you're raising in this case should be a `StudyImportError` with the custom message you wrote

tests/antares/services/api_services/test_study_api.py
```diff
@@ -713,3 +714,94 @@ def test_generate_thermal_timeseries_failure(self):
...
+
+    def test_import_study_fail_wrong_extension(self, tmp_path):
+        study_path = tmp_path.joinpath("test.rar")
+        study_path.touch()
+        new_path = Path("/new/path/test")
+
+        base_url = "https://antares.com/api/v1"
+        url_import = f"{base_url}/studies/_import"
+        with requests_mock.Mocker() as mocker:
+            mocker.get(url_import, status_code=404)
```
MartinBelthle, line 787: Don't think you need the mock inside this test as we raise before doing any request I believe

tests/antares/services/api_services/test_study_api.py
```diff
@@ -713,3 +714,94 @@ def test_generate_thermal_timeseries_failure(self):
...
+
+    def test_import_study_success(self, tmp_path):
+        json_study = {
+            "id": "22c52f44-4c2a-407b-862b-490887f93dd8",
+            "name": "test_read_areas",
+            "version": "880",
+            "folder": None,
+        }
+
+        json_ui = {
```
MartinBelthle, line 726: Don't think you need this. you could return an empty json I believe

tests/integration/test_web_client.py
```diff
@@ -92,6 +94,26 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop):
...
+        # importing without moving the study
+        imported_study = import_study_api(api_config, zip_study, None)
+
+        assert imported_study.path == study.path
+
+        # importing with moving the study
+        path_test = Path("/new/test/studies")
+        imported_study = import_study_api(api_config, zip_study, path_test)
+
+        assert imported_study.path == path_test / f"{imported_study.service.study_id}"
```
MartinBelthle, line 115: We could assert that the imported study has the same areas as the one you copied (to ensure the import went good)

MartinBelthle (changes requested):
Minor changes, otherwise looks good :)

src/antares/craft/model/study.py
```diff
@@ -89,6 +95,34 @@ def create_study_api(
...
+def import_study_api(api_config: APIconf, study_path: Path, destination_path: Optional[Path]) -> "Study":
+    def has_valid_extension(path: Path) -> bool:
+        valid_extensions = {".zip", ".7z"}
+        return path.suffix in valid_extensions
+
+    session = api_config.set_up_api_conf()
+    wrapper = RequestWrapper(session)
+    base_url = f"{api_config.get_host()}/api/v1"
+
+    if not has_valid_extension(study_path):
```
MartinBelthle, line 107: Could be simplified in 
`if study_path.suffix not in {".zip", ".7z"}` instead of introducing a function. It would be clearer for people reading the code

tests/antares/services/api_services/test_study_api.py
```diff
@@ -713,3 +714,80 @@ def test_generate_thermal_timeseries_failure(self):
...
+            mocker.get(renewable_url, json=[])
+            mocker.get(thermal_url, json=[])
+            mocker.get(storage_url, json=[])
+            mocker.get(
+                output_url,
+                json=[],
+            )
+            mocker.get(constraints_url, json=[])
+
+            # deux mockers
```
MartinBelthle, line 761: Remove this comment in French :)

tests/antares/services/api_services/test_study_api.py
```diff
@@ -713,3 +714,80 @@ def test_generate_thermal_timeseries_failure(self):
...
+
+            actual_study = import_study_api(self.api, study_path, new_path)
+
+            assert actual_study.name == json_study["name"]
+            assert actual_study.service.study_id == json_study["id"]
+
+    def test_import_study_fail_wrong_extension(self, tmp_path):
+        study_path = tmp_path.joinpath("test.rar")
+        study_path.touch()
+        new_path = Path("/new/path/test")
```
MartinBelthle, line 773: You don't need to introduce a destination_path here as your code will break before needing it.
Also you can just give a fake path (ex: `fake_path_for_test.rar`)without introducing a `tmp_path` as we only check the suffix not the existence

tests/antares/services/api_services/test_study_api.py
```diff
@@ -713,3 +714,80 @@ def test_generate_thermal_timeseries_failure(self):
...
+
+        with pytest.raises(Exception, match=re.escape("File doesn't have the right extensions (.zip/.7z): .rar")):
+            import_study_api(self.api, study_path, new_path)
+
+    def test_import_study_fail_api_error(self, tmp_path):
+        study_path = tmp_path.joinpath("test.zip")
```
MartinBelthle, line 779: Same remark here.
I don't think you need a `tmp_path` as you only check the suffix and then you're mocking the endpoints.

MartinBelthle (changes requested):
Small changes

MartinBelthle (approved):

PR #73, opened 2025-01-30
refactor(settings): refactor settings to create an user oriented class

MartinBelthle:


src/antares/craft/model/settings/study_settings.py
```diff
@@ -9,64 +9,23 @@
...
-        new_output = {key: value for key, value in output_dict["output"].items() if key != "result-format"} | {
-            "result-format": output_dict["output"]["result-format"]
-        }
-        output_dict["general"] = new_general
-        output_dict["output"] = new_output
-        return output_dict
+from antares.craft.model.settings.thematic_trimming import ThematicTrimmingParameters
+
+
+@dataclass
```
sylvlecl, line 23: We could make it frozen so that the user does not think that he can update the study simply by setting the properties on this object.

src/antares/craft/model/settings/study_settings.py
```diff
@@ -9,64 +9,23 @@
...
+class StudySettings:
+    general_parameters: Optional[GeneralParameters] = None
```
sylvlecl, line 25: In principle most of the fields should not be optional, I think.
When returning the settings of an existing study, we must not force the user to deal with optionals.

src/antares/craft/service/base_services.py
```diff
@@ -564,7 +564,7 @@ def config(self) -> BaseConfiguration:
         pass
 
     @abstractmethod
-    def update_study_settings(self, settings: StudySettings) -> Optional[StudySettings]:
+    def update_study_settings(self, settings: StudySettings) -> None:
```
sylvlecl, line 567: Here we could have a `StudySettignsUpdate` object with optionals.
But maybe you prefer to wait that we use the "function argument approach" before changing this ?
(where the update would take multiple named arguments)

src/antares/craft/tools/all_optional_meta.py
```diff
@@ -36,3 +37,17 @@ def all_optional_model(model: t.Type[ModelClass]) -> t.Type[ModelClass]:
         kwargs[field_name] = (new.annotation, new)
 
     return create_model(f"Partial{model.__name__}", __base__=model, __module__=model.__module__, **kwargs)  # type: ignore
+
+
+class LocalBaseModel(BaseModel):
```
sylvlecl, line 42: We should move it to local services module

src/antares/craft/model/study.py
```diff
@@ -169,20 +173,16 @@ def create_study_local(
...
         desktop_ini_file.write(desktop_ini_content)
 
-    local_settings = StudySettingsLocal.model_validate(settings)
-    local_settings_file = IniFile(study_directory, InitializationFilesTypes.GENERAL)
-    local_settings_file.ini_dict = local_settings.model_dump(exclude_none=True, by_alias=True)
-    local_settings_file.write_ini_file()
-
     # Create various .ini files for the study
-    _create_correlation_ini_files(local_settings, study_directory)
+    _create_correlation_ini_files(study_directory)
```
sylvlecl, line 177: Just a high level organization comment:
we should move the implementation of local and api methods to the corresponding services modules.
Then we can import them here so that it does not change the API. Just the code organization will be more clear.

sylvlecl (commented):
Looks good: still think that we should be more strict on the handling of optionals, removing them from where the user can expect to not have Nones.

sylvlecl (commented):
Also, I thinkw what we said for directory layout is:
- services
  - local
    - models
      - settings (all necessary classes)
    - services
      - settings (just the service class)

PR #74, opened 2025-02-04
Fix horizon to be accepted as str and read_study without renewables

vargastat:


tests/integration/test_web_client.py
```diff
@@ -658,3 +661,16 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop):
         moved_study = read_study_api(api_config, study.service.study_id)
         assert moved_study.path == study.path
         assert moved_study.name == study.name
+
+        new_settings_aggregated = StudySettings()
+        new_settings_aggregated.general_parameters = GeneralParameters(nbYears=4)
+        new_settings_aggregated.advanced_parameters = AdvancedParameters()
+        new_settings_aggregated.advanced_parameters.unit_commitment_mode = UnitCommitmentMode.MILP
+        new_settings_aggregated.advanced_parameters.renewable_generation_modelling = "aggregated"
+        study_aggregated = create_study_api("test_aggregated", "880", api_config, new_settings_aggregated)
```
MartinBelthle, line 670: You could just keep the line about aggregated to make the test clearer.
Also, the try catch is not necessary I think as you'll raise inside the function and so your test will fail

tests/antares/services/api_services/test_renewable_api.py
```diff
@@ -154,3 +154,23 @@ def test_read_renewables(self):
...
+                json={"description": "'renewables' not a child of Input", "exception": "ChildNotFoundError"},
+            )
+
+            renewable_api = RenewableApiService(self.api, study_id_test)
+
+            actual_renewable_list = renewable_api.read_renewables(area_id)
+
+            assert actual_renewable_list == []
+
+            assert len(actual_renewable_list) == 0
```
MartinBelthle, line 176: Don't need to check this as you've already checked it was an empty list right above

tests/antares/services/api_services/test_renewable_api.py
```diff
@@ -154,3 +154,23 @@ def test_read_renewables(self):
...
+        with requests_mock.Mocker() as mocker:
+            mocker.get(
+                url + "clusters/renewable",
+                status_code=404,
+                json={"description": "'renewables' not a child of Input", "exception": "ChildNotFoundError"},
+            )
+
+            renewable_api = RenewableApiService(self.api, study_id_test)
+
+            actual_renewable_list = renewable_api.read_renewables(area_id)
```
MartinBelthle, line 172: I think we should do `self.area.read_renewables()` instead, to fit with what an user would do

src/antares/craft/api_conf/request_wrapper.py
```diff
@@ -30,16 +30,29 @@
...
     """
     If an exception occurred, returns APIError exception containing the AntaresWeb error message.
     """
     if response.status_code - 200 < 100:
         return response
     try:
-        msg = response.json()["description"]
+        error_data = response.json()
+        msg = error_data.get("description", response.reason)
+        if response.status_code == 404 and "renewables" in msg:
```
MartinBelthle, line 51: I don't think you should introduce this here as this class is really general.
IMO it would be better to add a try catch inside the `read_renewables` method inside the API implem

MartinBelthle (changes requested):

tests/antares/services/api_services/test_renewable_api.py
vargastat: read_area has several mocks it is tested in integration test.

tests/antares/services/api_services/test_renewable_api.py
```diff
@@ -154,3 +154,21 @@ def test_read_renewables(self):
 
             assert expected_renewable.id == actual_renewable.id
             assert expected_renewable.name == actual_renewable.name
+
+    def test_read_renewables_with_404_response_then_return_empty_list(self):
```
MartinBelthle, line 158: I think this is a duplicate from the test inside `test_area_api.py` as I don't think we want to test the service directly (at least we didn't until now)

tests/integration/test_web_client.py
```diff
@@ -658,3 +661,14 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop):
...
         assert moved_study.path == study.path
         assert moved_study.name == study.name
+
+        new_settings_aggregated = StudySettings()
+        new_settings_aggregated.advanced_parameters = AdvancedParameters()
+        new_settings_aggregated.advanced_parameters.renewable_generation_modelling = "aggregated"
+        study_aggregated = create_study_api("test_aggregated", "880", api_config, new_settings_aggregated)
+        study_aggregated.create_area("area_without_renewables")
+        #  read_study_api does not raise an error
+        try:
```
MartinBelthle, line 671: I think you can remove the try / catch for readability sake, as the test will fail without it

src/antares/craft/service/api_services/renewable_api.py
```diff
@@ -83,8 +83,14 @@ def read_renewables(
         area_id: str,
     ) -> List[RenewableCluster]:
         url = f"{self._base_url}/studies/{self.study_id}/areas/{area_id}/clusters/renewable"
-        json_renewables = self._wrapper.get(url).json()
 
+        try:
+            json_renewables = self._wrapper.get(url).json()
+        except APIError as e:
+            if e.message == "'renewables' not a child of Input":
```
MartinBelthle, line 90: Could be nice to add a comment to specify why we're doing this weird thing

MartinBelthle (changes requested):

PR #79, opened 2025-02-11
refactor(thermal, renewable): create user classes

MartinBelthle:


src/antares/craft/service/api_services/area_api.py
```diff
@@ -148,63 +150,27 @@ def create_area(
...
         area_id: str,
         cluster_name: str,
-        parameters: ThermalClusterProperties,
+        properties: Optional[ThermalClusterProperties] = None,
         prepro: Optional[pd.DataFrame] = None,
         modulation: Optional[pd.DataFrame] = None,
         series: Optional[pd.DataFrame] = None,
-        CO2Cost: Optional[pd.DataFrame] = None,
-        fuelCost: Optional[pd.DataFrame] = None,
+        co2_cost: Optional[pd.DataFrame] = None,
```
sylvlecl, line 160: nice, the case was ugly

src/antares/craft/service/api_services/models/__init__.py
```diff
@@ -9,3 +9,9 @@
 # SPDX-License-Identifier: MPL-2.0
 #
 # This file is part of the Antares project.
+from pydantic import BaseModel, ConfigDict
+from pydantic.alias_generators import to_camel
+
+
+class APIBaseModel(BaseModel):
```
sylvlecl, line 16: Nice, maybe we could move it to a base_model.py module for more visibility

src/antares/craft/service/api_services/models/renewable.py
```diff
@@ -0,0 +1,47 @@
...
+    unit_count: int
+    nominal_capacity: float
+
+    @staticmethod
+    def from_user_model(user_class: RenewablePropertiesType) -> "RenewableClusterPropertiesAPI":
+        user_dict = asdict(user_class)
+        return RenewableClusterPropertiesAPI.model_validate(user_dict)
+
+    def to_user_model(self) -> RenewableClusterProperties:
+        return RenewableClusterProperties(
```
sylvlecl, line 41: It looks kind of dirty that it passes the type checking since strictly speaking the properties are optional now.
But we can dismiss that issue.

sylvlecl (approved):
Looks good, just non blocking minor comments

src/antares/craft/service/api_services/models/renewable.py
MartinBelthle: Indeed. We have the same issue concerning the settings. What should we do ? Introduce the same object with the default fields ?

src/antares/craft/service/api_services/models/__init__.py
MartinBelthle: Yes why not it's a bit hidden inside the init file

MartinBelthle: I'll do it inside another PR :)

src/antares/craft/service/api_services/models/renewable.py
MartinBelthle: Inside the PR concerning hydro, I did something to handle such case.
I do this:
```
default_properties = HydroProperties()
return HydroProperties(use_water=self.use_water or default_properties.use_water, ....)
```

PR #89, opened 2025-02-17
feat(api): add update thermal matrices methods

mehdiwahada:


MartinBelthle (changes requested):

src/antares/craft/service/api_services/services/thermal.py
```diff
@@ -67,14 +67,17 @@ def update_thermal_properties(
...
+        keyword = "series" if "SERIES" in ts_name.name else "prepro"
         path = (
             PurePosixPath("input")
             / "thermal"
-            / "series"
+            / keyword
             / f"{thermal_cluster.area_id}"
             / f"{thermal_cluster.id}"
-            / "series"
+            / ts_name.value
```
MartinBelthle, line 80: Line 85, we should add inside the exception the name of the matrix that couldn't be uploaded (data or modulation etc.)
You'll also have to add this inside your tests to make them work

MartinBelthle (changes requested):

MartinBelthle (approved):

PR #90, opened 2025-02-17
tests(hydro): add tests on matrices

mehdiwahada:


tests/antares/services/api_services/test_matrix_api.py
```diff
@@ -339,3 +403,19 @@ def test_get_credit_modulations_fail(self):
                 match=f"Error downloading credit_modulations matrix for area {self.area.id}: {self.antares_web_description_msg}",
             ):
                 self.hydro.get_credit_modulations()
+
+    def test_update_credit_modulations_success(self):
```
MartinBelthle, line 407: All these tests are good. We're just missing the same tests for 4 other matrices: `mod`, `ror`, `mingen` and `energy`
For these 4 matrices, we'll also have to do tests on the `GET`

tests/integration/test_web_client.py
```diff
@@ -567,6 +567,19 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop, tmp_path):
...
         actual_credit_matrix.equals(series)
 
+        # updating each hydro matrices
+        area_fr.hydro.update_maxpower(series)
+        area_fr.hydro.update_reservoir(series)
+        area_fr.hydro.update_inflow_pattern(series)
+        area_fr.hydro.update_water_values(series)
+        area_fr.hydro.update_credits_modulation(series)
+
+        actual_reservoir_matrix.equals(series)
```
MartinBelthle, line 577: You should do a get here as you're not checking the updated value but the old one you stored in the `GET`

MartinBelthle (changes requested):

tests/integration/test_web_client.py
```diff
@@ -560,12 +560,41 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop, tmp_path):
...
         actual_inflow_matrix.equals(series)
         actual_water_matrix.equals(series)
         actual_credit_matrix.equals(series)
+        actual_ror_matrix.equals(series)
+        actual_mod_matrix.equals(series)
+        actual_mingen_matrix.equals(series)
+        actual_energy_matrix.equals(series)
+
+        # updating each hydro matrices
+        area_fr.hydro.update_maxpower(series)
```
MartinBelthle, line 579: As I said in my last comment you should give another matrix than `series` here

MartinBelthle (changes requested):
Almost done :)

PR #97, opened 2025-02-25
Fixes/ matrix tool and area local service

killian-scalian:
- write_timeseries  now can be used to create series for path that have 2 patterns to match, and even if the path folders are missing
- series variable, if it is a dataframe, is not a condition enough accurate to know if it is empty or not

### DIFF
--- a/src/antares/craft/service/local_services/services/area.py
+++ b/src/antares/craft/service/local_services/services/area.py
@@ -155,9 +155,8 @@ class AreaLocalService(BaseAreaService):
         list_ini = IniFile(self.config.study_path, InitializationFilesTypes.RENEWABLES_LIST_INI, area_id=area_id)
         list_ini.add_section({renewable_name: new_section_content})
         list_ini.write_ini_file()
-
-        if series:
-            write_timeseries(self.config.study_path, series, TimeSeriesFileType.RENEWABLE_DATA_SERIES, area_id)
+        if isinstance(series, pd.DataFrame) and not series.empty:
+            write_timeseries(self.config.study_path, series, TimeSeriesFileType.RENEWABLE_DATA_SERIES, area_id, cluster_id=transform_name_to_id(renewable_name))
 
         return RenewableCluster(self.renewable_service, area_id, renewable_name, properties)
 
--- a/src/antares/craft/tools/matrix_tool.py
+++ b/src/antares/craft/tools/matrix_tool.py
@@ -65,6 +65,13 @@ def read_timeseries(
     return _time_series
 
 
-def write_timeseries(study_path: Path, series: pd.DataFrame, ts_file_type: TimeSeriesFileType, area_id: str) -> None:
-    file_path = study_path.joinpath(ts_file_type.value.format(area_id=area_id))
+def write_timeseries(study_path: Path, series: pd.DataFrame, ts_file_type: TimeSeriesFileType, area_id: str, cluster_id: Optional[str] = None, second_area_id: Optional[str] = None) -> None:
+    if second_area_id:
+        file_path = study_path.joinpath(ts_file_type.value.format(area_id=area_id, second_area_id=second_area_id))
+    elif cluster_id:
+        file_path = study_path.joinpath(ts_file_type.value.format(area_id=area_id, cluster_id=cluster_id))
+    else:
+        file_path = study_path.joinpath(ts_file_type.value.format(area_id=area_id))
+    if not file_path.parent.is_dir():
+        file_path.parent.mkdir(parents=True)
     series.to_csv(file_path, sep="\t", header=False, index=False, encoding="utf-8")


src/antares/craft/service/local_services/services/area.py
```diff
@@ -155,9 +155,14 @@ def create_renewable_cluster(
         list_ini = IniFile(self.config.study_path, InitializationFilesTypes.RENEWABLES_LIST_INI, area_id=area_id)
         list_ini.add_section({renewable_name: new_section_content})
         list_ini.write_ini_file()
-
-        if series:
-            write_timeseries(self.config.study_path, series, TimeSeriesFileType.RENEWABLE_DATA_SERIES, area_id)
+        if isinstance(series, pd.DataFrame) and not series.empty:
```
MartinBelthle, line 158: The behavior we want here is to create an empty file if there's no series or if the df is empty.

So I think what you should do instead is:
`series = series or pd.DataFrame()`

and if needed change the `write_timeseries` method to do a `file_path.touch()` if the df is empty

src/antares/craft/tools/matrix_tool.py
```diff
@@ -65,6 +65,27 @@ def read_timeseries(
...
+def write_timeseries(
+    study_path: Path,
+    series: pd.DataFrame,
+    ts_file_type: TimeSeriesFileType,
+    area_id: str,
+    cluster_id: Optional[str] = None,
+    second_area_id: Optional[str] = None,
+    constraint_id: Optional[str] = None,
+) -> None:
+    # This method does not concern binding constraints timeseries, which is handled differently
```
MartinBelthle, line 77: It seems that this method handles constraints as there's an arg named `constraint_id`

tests/antares/tools/test_contents_tool.py
```diff
@@ -31,3 +34,21 @@ def test_prepare_args():
 
     # Then
     assert result == expected_result
+
+
+def test_write_timeseries(tmpdir):
+    file_path = Path(tmpdir)
+    df = pd.DataFrame([1, 2, 3], columns=["Value"])
+
+    write_timeseries(file_path, df, TimeSeriesFileType.THERMAL_MODULATION, area_id="fr", cluster_id="gaz")
```
MartinBelthle, line 43: You could add a test for constraint_id and for empty df (check with pathlib `read_text()` method)

MartinBelthle (changes requested):

src/antares/craft/tools/matrix_tool.py
killian-scalian: I forgot to remove it

killian-scalian: I should have remove it indeed. i will next commit

killian-scalian: After discussion i implemented BC with timeseries

src/antares/craft/service/local_services/services/area.py
killian-scalian: Done in the next commit

tests/antares/services/local_services/conftest.py
```diff
@@ -133,7 +133,16 @@ def actual_adequacy_patch_ini(local_study_w_areas) -> IniFile:
 def local_study_with_renewable(local_study_w_thermal) -> Study:
     renewable_cluster_name = "renewable cluster"
     local_study_w_thermal.get_areas()["fr"].create_renewable_cluster(
-        renewable_cluster_name, RenewableClusterProperties(), None
+        renewable_cluster_name, RenewableClusterProperties(), series=None
+    )
+    return local_study_w_thermal
+
+
+@pytest.fixture
```
MartinBelthle, line 141: Not sure you need to introduce a fixture for that.
You can just do this in your test it will be clearer

tests/antares/services/local_services/test_area.py
```diff
@@ -116,6 +116,9 @@ def test_renewable_cluster_and_ini_have_custom_properties(self, local_study_w_th
         )
         assert actual_renewable_list_ini_content == expected_renewables_list_ini_content
 
+    def test_renewable_cluster_and_series_is_empty(self, local_study_with_renewable_and_empty_dataframe):
+        assert local_study_with_renewable_and_empty_dataframe.service.config.study_path.exists()
```
MartinBelthle, line 120: You can also check that this file exists even when the series is not given

MartinBelthle: And that it's empty

MartinBelthle (changes requested):

src/antares/craft/service/local_services/services/area.py
```diff
@@ -155,10 +155,13 @@ def create_renewable_cluster(
         list_ini = IniFile(self.config.study_path, InitializationFilesTypes.RENEWABLES_LIST_INI, area_id=area_id)
         list_ini.add_section({renewable_name: new_section_content})
         list_ini.write_ini_file()
-
```
MartinBelthle, line 158: I also realized that from line 130 to 141 we need to remove the `if` and also add the arg cluster_id when calling the method

MartinBelthle (commented):
Minor change not really related to the PR

MartinBelthle (approved):

PR #108, opened 2025-03-04
feat(api): introduce `update_multiple_constraints` method

mehdiwahada:


src/antares/craft/model/study.py
```diff
@@ -251,6 +252,13 @@ def delete_binding_constraint(self, constraint: BindingConstraint) -> None:
         self._study_service.delete_binding_constraint(constraint)
         self._binding_constraints.pop(constraint.id)
 
+    def update_multiple_binding_constraints(
+        self, b_constraint_update: Dict[str, "BindingConstraintPropertiesUpdate"]
```
MartinBelthle, line 256: You can remove the quotes around `BindingConstraintPropertiesUpdate`.
Also, the name `b_constraint_update` is not really clear, you can rename it `new_properties` as you did for links update

src/antares/craft/service/api_services/services/study.py
```diff
@@ -145,3 +151,31 @@ def generate_thermal_timeseries(self, nb_years: int) -> None:
             wait_task_completion(self._base_url, self._wrapper, task_id)
         except (APIError, TaskFailedError, TaskTimeOutError) as e:
             raise ThermalTimeseriesGenerationError(self.study_id, e.message)
+
+    @override
+    def update_multiple_binding_constraints(
+        self, b_constraint_update: Dict[str, "BindingConstraintPropertiesUpdate"]
```
MartinBelthle, line 157: Same, you can remove the quotes

src/antares/craft/service/api_services/services/study.py
```diff
@@ -145,3 +151,31 @@ def generate_thermal_timeseries(self, nb_years: int) -> None:
...
+            body[bc_id] = api_dict
+
+        try:
+            binding_constraints_dict = self._wrapper.put(url, json=body).json()
+
+            for binding_constraints in binding_constraints_dict:
+                api_response = BindingConstraintPropertiesAPI.model_validate(
+                    binding_constraints_dict[binding_constraints]
+                )
+                constraints_properties = api_response.model_dump(mode="json", by_alias=True, exclude_none=True)
```
MartinBelthle, line 175: This doesn't work. You need to convert the `api_response` to a BindingConstraintProperties object.
You can do this like that `api_response.to_user_model()`

src/antares/craft/service/local_services/services/study.py
```diff
@@ -73,3 +77,9 @@ def move_study(self, new_parent_path: Path) -> PurePath:
     @override
     def generate_thermal_timeseries(self, number_of_years: int) -> None:
         raise NotImplementedError
+
+    @override
+    def update_multiple_binding_constraints(
+        self, b_constraint_update: Dict[str, "BindingConstraintPropertiesUpdate"]
```
MartinBelthle, line 83: Minor: you can change `Dict` to `dict` to avoid importing it from typing

tests/integration/test_web_client.py
```diff
@@ -373,6 +373,24 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop, tmp_path):
...
+        # test updating constraints
+        update_bc_props_1 = BindingConstraintPropertiesUpdate(time_step=BindingConstraintFrequency.DAILY, enabled=False)
+        update_bc_props_2 = BindingConstraintPropertiesUpdate(
+            enabled=True, time_step=BindingConstraintFrequency.HOURLY, comments="Bonjour"
+        )
+        dict_binding_constraints_update = {constraint_1.name: update_bc_props_1, constraint_2.name: update_bc_props_2}
+
+        study.update_multiple_binding_constraints(dict_binding_constraints_update)
+
+        constraint_1 = study._binding_constraints["bc_1"]
```
MartinBelthle, line 385: I did a dev in the mean time that should allow you to just check with `constraint_1` instead of redeclaring it

tests/integration/test_web_client.py
```diff
@@ -373,6 +373,24 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop, tmp_path):
...
+        constraint_2 = study._binding_constraints["bc_2"]
+
+        assert constraint_1._properties["timeStep"] == "daily"
```
MartinBelthle, line 388: You shouldn't access private fields like this, you should just use the property `properties`. Also this won't be a dict anymore so you should check with `.time_step`

tests/integration/test_web_client.py
```diff
@@ -373,6 +373,24 @@ def test_creation_lifecycle(self, antares_web: AntaresWebDesktop, tmp_path):
...
+        assert not constraint_1._properties["enabled"]
+        assert constraint_2._properties["timeStep"] == "hourly"
+        assert constraint_2._properties["enabled"]
+        assert constraint_2._properties["comments"] == "Bonjour"
```
MartinBelthle, line 392: You can add a test to ensure an old properties of constraint_1 was not modified

tests/antares/services/api_services/test_study_api.py
```diff
@@ -904,3 +910,44 @@ def test_update_multiple_links_fail(self):
...
+            dict_binding_constraints.update({bc_id: bc_props})
+
+        with requests_mock.Mocker() as mocker:
+            mocker.put(url, json=json_binding_constraints)
+            self.study.update_multiple_binding_constraints(dict_binding_constraints)
+
+            bc_test_1 = self.study._binding_constraints["battery_state_evolution"]
+            bc_test_2 = self.study._binding_constraints["battery_state_update"]
+
+            assert bc_test_1._properties["enabled"] == dict_binding_constraints["battery_state_evolution"].enabled
```
MartinBelthle, line 937: Same comments as the ones inside `test_web_client`

MartinBelthle (changes requested):

PR #121, opened 2025-03-10
doc: initialize doc deployment to readthedoc

sylvlecl:
The goal of this PR is to setup the main structure around the documentation of the library on readthedocs.
It contains:
- an updated README
- configuration for readthedoc and mkdocs
- main layout of the doc, split into: getting started, user guide, API reference, and developers section.
- a few improvements/updates of the existing documentation
- generation of the API reference from the python code of `antares.craft` module

We will then need to work on the actual content.

The current generated documentation can be found here:
https://antares-craft.readthedocs.io/en/feat-readthedoc/getting-started/


MartinBelthle (changes requested):
You broke the README.md concatenation which is performed on release for Pypi.
It's because you removed the usage.md file. I'll let you change the script Tatiana wrote (`/scripts/concat_readme.md`) :)

MartinBelthle:
Same remark, inside the README.md we point towards this removed file so we'll have to change that

MartinBelthle (approved):

PR #139, opened 2025-03-18
Accept native bools for ExportMPS enum class

flomnes:
It becomes possible to write
```py
craft.ExportMPS(True)
craft.ExportMPS(False)
```
instead of
```
craft.ExportMPS("True")
craft.ExportMPS("False")
```

I added some unit tests to check the associated behavior.


MartinBelthle:
I do not quite understand the purpose of the PR. As an user I would write ExportMPS.FALSE and I wouldn't check what's the associated value.
So I'm probably missing something, feel free to explain inside the PR doc :)

PR #149, opened 2025-03-24
feat(renewable): add `update_renewable_clusters` method

mehdiwahada:


src/antares/craft/exceptions/exceptions.py
```diff
@@ -180,6 +180,12 @@ def __init__(self, area_id: str, renewable_names: List[str], message: str) -> No
         super().__init__(self.message)
 
 
+class RenewablesPropertiesUpdateError(Exception):
+    def __init__(self, study_id: str, message: str):
+        self.message = f"Could not update properties of the clusters from study {study_id} : {message}"
```
MartinBelthle, line 185: Could be nice to add `renewable` before clusters

src/antares/craft/model/study.py
```diff
@@ -304,6 +305,17 @@ def update_thermal_clusters(self, new_properties: dict[ThermalCluster, ThermalCl
...
             self._areas[thermal.area_id].get_thermals()[thermal.id]._properties = new_thermal_clusters_props[thermal]
 
+    def update_renewable_clusters(
+        self, new_properties: dict[RenewableCluster, RenewableClusterPropertiesUpdate]
+    ) -> None:
+        new_renewable_clusters_props = self._area_service.renewable_service.update_renewable_clusters_properties(
+            new_properties
+        )
+        for renewable in new_renewable_clusters_props:
+            self._areas[renewable.area_id].get_renewables()[renewable.id]._properties = new_renewable_clusters_props[
```
MartinBelthle, line 315: I think you should'nt use `get_renewables()` but `_renewables` as the get method returns a copy of the dict

tests/antares/services/api_services/test_renewable_api.py
```diff
@@ -42,7 +45,17 @@ class TestCreateAPI:
         services.renewable_service,
         services.hydro_service,
     )
+    area_1 = Area(
```
MartinBelthle, line 48: Don't think we need to introduce this area, same for renewable_2. 
An you could put renewable_1 inside the existing area

tests/antares/services/api_services/test_renewable_api.py
```diff
@@ -156,3 +169,81 @@ def test_read_renewables(self):
...
+                updated_renewable[renewable] = renewable_update
+
+            mocker.put(url, json=json_renewables)
+
+            self.study.update_renewable_clusters(updated_renewable)
+
+            renewable_1 = self.study._areas["at"]._renewables["at_solar_pv"]
+            renewable_2 = self.study._areas["at"]._renewables["at_solar_thermo"]
+
+            print(renewable_1.properties)
```
MartinBelthle, line 228: You can remove these `print`

MartinBelthle (changes requested):
You can also implement the local part. You can see what's been done for thermals

PR #174, opened 2025-04-16
ci: speed-up windows ci

MartinBelthle:


MartinBelthle:
Doesn't speed-up the perfs

PR #179, opened 2025-04-18
feat(version): handle Simulator v9.2

berthetclement:
**TODO**:

- Additional constraints (local + API), only in AWeb 2.23
- New simulation parameters (local + API), only in AWeb 2.23


src/antares/craft/model/st_storage.py
vargastat, line 1: parameters  penalize-variation-injection boolean, penalize-variation-withdrawal to be added

src/antares/craft/service/local_services/factory.py
```diff
@@ -37,13 +37,13 @@
 from antares.craft.tools.serde_local.ini_writer import IniWriter
 
 
-def create_local_services(config: LocalConfiguration, study_name: str = "") -> StudyServices:
+def create_local_services(config: LocalConfiguration, study_name: str = "", study_version: str = "") -> StudyServices:
```
MartinBelthle, line 40: We should rather put the study_version inside the LocalConfiguration object. I'll do a PR to do this and modify a bit the study creation to use another Python package we're already maintaining `antares-study-version`

src/antares/craft/service/local_services/services/area.py
```diff
@@ -215,15 +217,34 @@ def set_load(self, area_id: str, series: pd.DataFrame) -> None:
...
         properties = properties or STStorageProperties()
         local_properties = STStoragePropertiesLocal.from_user_model(properties)
 
         local_storage_service = cast(ShortTermStorageLocalService, self.storage_service)
         ini_content = local_storage_service.read_ini(area_id)
-        ini_content[st_storage_name] = {
-            "name": st_storage_name,
-            **local_properties.model_dump(mode="json", by_alias=True),
-        }
+        if self.study_version == "880":
```
MartinBelthle, line 235: This distinction should be done inside the class `STStoragePropertiesLocal`

MartinBelthle (changes requested):
We're missing several changes that the v9.2 brings (see the Simulator documentation):
- 5 cost matrices for the STS
- A new field for hydro
- Several fields disappeared and another was created inside the generaldata)

src/antares/craft/service/local_services/factory.py
MartinBelthle: After reading the code, I'm not sure we should do what I said. I think we'll have to pass the study_version inside the services. But now the study version is a StudyVersion object you'll have to change the typing

docs/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+v0.2.6.x (2025-04-17)
```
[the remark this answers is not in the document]
MartinBelthle: Same remark

src/antares/craft/model/st_storage.py
```diff
@@ -42,26 +42,33 @@ class STStorageMatrixName(Enum):
 
 @dataclass
 class STStoragePropertiesUpdate:
-    group: Optional[STStorageGroup] = None
+    group: Optional[Union[STStorageGroup, str]] = None
```
MartinBelthle, line 45: Since Python 3.10 `Union` is deprecated we tend to use `|` instead which does the same thing

src/antares/craft/model/st_storage.py
```diff
@@ -42,26 +42,33 @@ class STStorageMatrixName(Enum):
...
     enabled: Optional[bool] = None
+    efficiency_withdrawal: Optional[float] = None
+    penalize_variation_injection: Optional[bool] = None
+    penalize_variation_withdrawal: Optional[bool] = None
 
 
 @dataclass(frozen=True)
 class STStorageProperties:
-    group: STStorageGroup = STStorageGroup.OTHER1
+    group: Optional[STStorageGroup | str] = None
```
MartinBelthle, line 60: I think default value shouldn't be none but `STStorageGroup.OTHER1`

src/antares/craft/service/api_services/models/st_storage.py
```diff
@@ -28,10 +30,14 @@ class STStoragePropertiesAPI(APIBaseModel):
     initial_level: float
     initial_level_optim: bool
     enabled: bool
+    # add new parameter 9.2 but not study_version validation in API
+    efficiency_withdrawal: float = 1
+    penalize_variation_injection: bool = Field(False, serialization_alias="penalize-variation-injection")
```
MartinBelthle, line 35: You don't have to use Field and you should remove the serialization alais as we use camel case for API
Also you don't have to put a default value as it will be overwritten by None with the decorator

src/antares/craft/service/api_services/models/st_storage.py
```diff
@@ -28,10 +30,14 @@ class STStoragePropertiesAPI(APIBaseModel):
...
+    penalize_variation_withdrawal: bool = Field(False, serialization_alias="penalize-variation-withdrawal")
 
     @staticmethod
     def from_user_model(user_class: STStoragePropertiesType) -> "STStoragePropertiesAPI":
-        user_dict = asdict(user_class)
+        user_dict = {k: v for k, v in asdict(user_class).items() if v is not None}
```
MartinBelthle, line 40: Don't think we need to do this

src/antares/craft/service/local_services/factory.py
```diff
@@ -35,21 +35,27 @@
 from antares.study.version.create_app import CreateApp
 
 
-def create_local_services(config: LocalConfiguration, study_name: str = "") -> StudyServices:
-    storage_service = ShortTermStorageLocalService(config, study_name)
+def create_local_services(config: LocalConfiguration, study_name: str = "", study_version: str = "") -> StudyServices:
```
MartinBelthle, line 38: Here we should give the study_version as a StudyVersion object already

src/antares/craft/service/local_services/models/st_storage.py
```diff
@@ -22,28 +27,74 @@ def _sts_alias_generator(input: str) -> str:
...
     enabled: bool = True
+    # add new parameter 9.2
+    efficiency_withdrawal: float = 1
+
+    penalize_variation_injection: bool = Field(False, alias="penalize-variation-injection")
+    penalize_variation_withdrawal: bool = Field(False, alias="penalize-variation-withdrawal")
+
+    @field_validator("group")
+    def validate_group(cls, v: str | STStorageGroup | None, info: ValidationInfo) -> str | STStorageGroup:
+        study_version = getattr(info.context, "get", lambda x: None)("study_version")
```
MartinBelthle, line 46: Why not just `info.context["study_version"]` ?

src/antares/craft/service/local_services/models/st_storage.py
```diff
@@ -22,28 +27,74 @@ def _sts_alias_generator(input: str) -> str:
...
+                try:
+                    return STStorageGroup(v)
+                except ValueError:
+                    raise ValueError(f"Group for 8.8 has to be a valid value : {[e.value for e in STStorageGroup]}")
+            return v
+        return v if v is not None else ""
+
+    @field_validator("efficiency_withdrawal", "penalize_variation_injection", "penalize_variation_withdrawal")
+    def check_version_support(cls, v: Union[str, STStorageGroup], info: ValidationInfo) -> Any:
+        study_version = getattr(info.context, "get", lambda x: None)("study_version")
```
MartinBelthle, line 60: Same

src/antares/craft/service/local_services/models/st_storage.py
```diff
@@ -22,28 +27,74 @@ def _sts_alias_generator(input: str) -> str:
...
+        if study_version == StudyVersion.parse("8.8"):
+            return None
```
MartinBelthle, line 62: We shouldn't just put None but rather crash if there are 9.2 fields inside a 8.8 study

src/antares/craft/service/local_services/models/st_storage.py
```diff
@@ -22,28 +27,74 @@ def _sts_alias_generator(input: str) -> str:
...
+        return v
 
     @staticmethod
-    def from_user_model(user_class: STStoragePropertiesType) -> "STStoragePropertiesLocal":
+    def from_user_model(user_class: STStoragePropertiesType, study_version: StudyVersion) -> "STStoragePropertiesLocal":
         user_dict = {k: v for k, v in asdict(user_class).items() if v is not None}
-        return STStoragePropertiesLocal.model_validate(user_dict)
+
+        if study_version == StudyVersion.parse("8.8"):
```
MartinBelthle, line 69: We should use the constant `STUDY_VERSION_8_8` instead of doing this everywhere

src/antares/craft/model/st_storage.py
MartinBelthle: Also, after discussion with Sylvain, we decided it would be better to type it only as a string but keep the validation with `STStorageGroup(value)` inside the implems to ensure the user respected the group

src/antares/craft/service/local_services/models/st_storage.py
```diff
@@ -22,28 +27,74 @@ def _sts_alias_generator(input: str) -> str:
...
+                except ValueError:
+                    raise ValueError(f"Group for 8.8 has to be a valid value : {[e.value for e in STStorageGroup]}")
+        else:
+            user_dict["group"] = user_dict.get("group", "")
+
+        return STStoragePropertiesLocal.model_validate(user_dict, context={"study_version": study_version})
 
     def to_user_model(self) -> STStorageProperties:
+        group_value = self.group
+        if hasattr(self, "_context") and self._context.get("study_version") == StudyVersion.parse("8.8"):
```
MartinBelthle, line 84: We never go through this code

src/antares/craft/model/st_storage.py
```diff
@@ -42,26 +42,33 @@ class STStorageMatrixName(Enum):
...
     injection_nominal_capacity: float = 0
     withdrawal_nominal_capacity: float = 0
     reservoir_capacity: float = 0
     efficiency: float = 1
     initial_level: float = 0.5
     initial_level_optim: bool = False
     enabled: bool = True
+    # add new parameter 9.2
+    efficiency_withdrawal: float = 1
```
MartinBelthle, line 69: These should be None by default

MartinBelthle (changes requested):

src/antares/craft/service/local_services/models/st_storage.py
vargastat: it was to avoid bad idea to avoid None et mypy issue

src/antares/craft/model/st_storage.py
berthetclement: fix with `STStorageGroup.OTHER1` (default value like 8.8) + update test `test_storage_group_version_handling`

MartinBelthle: This is not relevant anymore due to previous comment :/
cf. `Also, after discussion with Sylvain, we decided it would be better to type it only as a string but keep the validation with STStorageGroup(value) inside the implems to ensure the user respected the group`

src/antares/craft/model/st_storage.py
vargastat: ok it is now str and validation done at implementation using STStorageGroup(value) for 8.8

src/antares/craft/model/st_storage.py
vargastat: Yes , agree it is better str, for 9.2 I leave an empty string if nothing provided, unit test added if we change our mind

src/antares/craft/model/st_storage.py
```diff
@@ -42,26 +42,33 @@ class STStorageMatrixName(Enum):
...
     enabled: Optional[bool] = None
+    efficiency_withdrawal: Optional[float] = None
+    penalize_variation_injection: Optional[bool] = None
+    penalize_variation_withdrawal: Optional[bool] = None
 
 
 @dataclass(frozen=True)
 class STStorageProperties:
-    group: STStorageGroup = STStorageGroup.OTHER1
+    group: str = ""
```
MartinBelthle, line 60: I think it should be other 1 by default as otherwise the cast to StStorageGroup will fail when we're in v8.8
Also it's explicit for the user that it's this value that we'll take

src/antares/craft/service/local_services/factory.py
```diff
@@ -35,21 +35,27 @@
...
+def create_local_services(config: LocalConfiguration, study_name: str, study_version: StudyVersion) -> StudyServices:
+    storage_service = ShortTermStorageLocalService(config, study_name, study_version)
     thermal_service = ThermalLocalService(config, study_name)
     renewable_service = RenewableLocalService(config, study_name)
     hydro_service = HydroLocalService(config, study_name)
     area_service = AreaLocalService(
-        config, study_name, storage_service, thermal_service, renewable_service, hydro_service
+        config,
+        study_name,
+        StudyVersion.parse(study_version),
```
MartinBelthle, line 46: No need to parse as we already provide a StudyVersion object
Same goes for other parts of this file

src/antares/craft/service/local_services/models/st_storage.py
```diff
@@ -22,28 +28,85 @@ def _sts_alias_generator(input: str) -> str:
 
 
 class STStoragePropertiesLocal(LocalBaseModel, alias_generator=_sts_alias_generator):
-    group: STStorageGroup = STStorageGroup.OTHER1
+    group: str = ""
```
MartinBelthle, line 31: Same here concerning other 1

src/antares/craft/service/local_services/models/st_storage.py
```diff
@@ -22,28 +28,85 @@ def _sts_alias_generator(input: str) -> str:
...
+                raise ValueError(f"In version 8.8, the following values are not allowed: {', '.join(violations)}")
+
+        return STStoragePropertiesLocal.model_validate(user_dict, context={"study_version": study_version})
 
     def to_user_model(self) -> STStorageProperties:
+        group_value = self.group
         return STStorageProperties(
             enabled=self.enabled,
-            group=self.group,
+            group=group_value,
```
MartinBelthle, line 102: You can just keep self.group

src/antares/craft/service/local_services/models/st_storage.py
```diff
@@ -22,28 +28,85 @@ def _sts_alias_generator(input: str) -> str:
...
+                except ValueError:
+                    raise ValueError(f"Group for 8.8 has to be a valid value : {[e.value for e in STStorageGroup]}")
+            if isinstance(v, STStorageGroup):
+                return v.value
+        return str(v) if v is not None else ""
+
+    @field_validator("efficiency_withdrawal", "penalize_variation_injection", "penalize_variation_withdrawal")
+    def check_version_support(cls, v: Union[str, STStorageGroup], info: ValidationInfo) -> Any:
+        study_version = info.context.get("study_version") if info.context is not None else None
+        if study_version == STUDY_VERSION_8_8:
```
MartinBelthle, line 64: We should raise if 9.2 fields are given for v8.8 study and here I don't think we do

src/antares/craft/service/local_services/models/st_storage.py
```diff
@@ -22,28 +28,85 @@ def _sts_alias_generator(input: str) -> str:
...
+            group = user_dict.get("group")
+            if group is None or group == "":
+                user_dict["group"] = STStorageGroup.OTHER1.value
+                group = user_dict["group"]
+            valid_values = [e.value for e in STStorageGroup]
+            if group not in valid_values:
+                raise ValueError("In 8.8, group must be a valid string value from STStorageGroup")
+
+        if study_version == STUDY_VERSION_8_8 and custom_properties is True:
```
MartinBelthle, line 83: The arg `custom_properties` should be removed and we could simplify the check by just writing 
```python
violations = [field if field in user_dict for field in field_to_check]
```

src/antares/craft/service/local_services/models/st_storage.py
```diff
@@ -22,28 +28,85 @@ def _sts_alias_generator(input: str) -> str:
...
+                    return v
+                except ValueError:
+                    raise ValueError(f"Group for 8.8 has to be a valid value : {[e.value for e in STStorageGroup]}")
+            if isinstance(v, STStorageGroup):
+                return v.value
+        return str(v) if v is not None else ""
+
+    @field_validator("efficiency_withdrawal", "penalize_variation_injection", "penalize_variation_withdrawal")
+    def check_version_support(cls, v: Union[str, STStorageGroup], info: ValidationInfo) -> Any:
+        study_version = info.context.get("study_version") if info.context is not None else None
```
MartinBelthle, line 63: This is really hard to read. I think we should encapsulate the model_validate inside a `read_ini` method where we would give the study_version and do the verficiation and at the end just do the model validate.

With this change we won't have to do multiple verification when doing a from_user_model as the current code currently does

MartinBelthle (changes requested):

src/antares/craft/service/local_services/models/st_storage.py
vargastat: We keep for a while as it ensures case when not custom properties are provided

src/antares/craft/service/local_services/models/st_storage.py
vargastat: As for @dataClass if we keep default values we need to keep custom_properties to check if it has changed then raise exception..maybe it can be changed if we put Optional in model definition

src/antares/craft/service/local_services/factory.py
vargastat: done in good places

src/antares/craft/model/st_storage.py
```diff
@@ -37,30 +37,43 @@ class STStorageMatrixName(Enum):
...
     injection_nominal_capacity: float = 0
     withdrawal_nominal_capacity: float = 0
     reservoir_capacity: float = 0
     efficiency: float = 1
     initial_level: float = 0.5
     initial_level_optim: bool = False
     enabled: bool = True
+    # add new parameter 9.2
+    efficiency_withdrawal: Optional[float] = None
+    penalize_variation_injection: Optional[float] = None
```
MartinBelthle, line 75: The 2 `penalize` parameter are bools, I'm surprised mypy didn't say a thing
Also we should add a comment saying what will be written values even if it's written None for the user to understand

src/antares/craft/service/local_services/factory.py
```diff
@@ -133,7 +139,9 @@ def read_study_local(study_directory: Path, solver_path: Optional[Path] = None)
     study = Study(
         name=study_params["caption"],
         version=study_params["version"],
-        services=create_local_services(config=local_config, study_name=study_params["caption"]),
+        services=create_local_services(
+            config=local_config, study_name=study_params["caption"], study_version=study_params["version"]
```
MartinBelthle, line 143: Here you should parse the version

tests/antares/services/api_services/test_area_api.py
```diff
@@ -178,7 +178,11 @@ def test_create_renewable_fails(self) -> None:
     def test_create_st_storage_success(self) -> None:
         with requests_mock.Mocker() as mocker:
             url = f"{self.study_url}/areas/{self.area.id}/storages"
-            json_response = STStoragePropertiesAPI(**{}).model_dump(mode="json", by_alias=True)
+            json_response = STStoragePropertiesAPI(**{}).model_dump(
+                mode="json",
+                by_alias=True,
+                exclude={"efficiency_withdrawal", "penalize_variation_injection", "penalize_variation_withdrawal"},
```
MartinBelthle, line 184: Put a TODO here to say we'll have to remove it when AntaresWbe supports the 9.2

tests/antares/services/local_services/test_area.py
```diff
@@ -132,13 +133,17 @@ def test_storage_has_correct_default_properties(self, local_study_with_st_storag
         st_storage = local_study_with_st_storage.get_areas()["fr"].get_st_storages()["short term storage"]
         assert st_storage.properties == STStorageProperties()
 
+    def test_storage_has_correct_default_properties_92(self, local_study_92: Study) -> None:
+        st_storage = local_study_92.get_areas()["fr"].create_st_storage("short term storage")
+        assert st_storage.properties == STStorageProperties()
```
MartinBelthle, line 138: Seems that we shouldn't have this but rather an object with the 9.2 properties filled

src/antares/craft/service/local_services/services/area.py
```diff
@@ -214,15 +218,28 @@ def set_load(self, area_id: str, series: pd.DataFrame) -> None:
     def create_st_storage(
         self, area_id: str, st_storage_name: str, properties: Optional[STStorageProperties] = None
     ) -> STStorage:
+        """
```
MartinBelthle, line 221: We should move this doc inside the base service

src/antares/craft/service/local_services/services/area.py
```diff
@@ -237,6 +254,7 @@ def create_st_storage(
         cluster_id = storage.id
         default_matrix_ones = pd.DataFrame(default_series_with_ones)
         default_matrix_zeros = pd.DataFrame(default_series)
+        empty_matrix = pd.DataFrame()
         # fmt: off
         write_timeseries(self.config.study_path, default_matrix_ones, TimeSeriesFileType.ST_STORAGE_PMAX_INJECTION, area_id, cluster_id=cluster_id)
```
MartinBelthle, line 259: We no longer have to write the default matrices since simulator v8.8.15 so we can just use empty_matrix here instead of the default matrices

src/antares/craft/service/local_services/models/st_storage.py
```diff
@@ -22,28 +28,107 @@ def _sts_alias_generator(input: str) -> str:
...
+
+    @field_validator("group")
+    def validate_group(cls, v: str | STStorageGroup | None, info: ValidationInfo) -> str:
+        study_version = info.context.get("study_version") if info.context is not None else None
+        if study_version == STUDY_VERSION_8_8:
+            if v is None or v == "":
+                return STStorageGroup.OTHER1.value
+            if isinstance(v, str):
+                try:
+                    _ = STStorageGroup(v)
```
MartinBelthle, line 52: Theo made a remark in AntaresWeb it would probably be easier to read to write directly 
`if v not in [e.value for e in STStorageGroup]: raise ...`

src/antares/craft/service/local_services/models/st_storage.py
```diff
@@ -22,28 +28,107 @@ def _sts_alias_generator(input: str) -> str:
...
-            group=self.group,
-            injection_nominal_capacity=self.injection_nominal_capacity,
-            withdrawal_nominal_capacity=self.withdrawal_nominal_capacity,
-            reservoir_capacity=self.reservoir_capacity,
-            efficiency=self.efficiency,
-            initial_level=self.initial_level,
-            initial_level_optim=self.initial_level_optim,
-        )
+        # For version 9.2 and above, include the additional properties
+        if self.efficiency_withdrawal is not None:
```
MartinBelthle, line 109: This check is really weird, it would be better to give the version inside the `to_user_model` if possible

src/antares/craft/service/local_services/models/st_storage.py
```diff
@@ -22,28 +28,107 @@ def _sts_alias_generator(input: str) -> str:
...
+            if group not in valid_values:
+                raise ValueError("In 8.8, group must be a valid string value from STStorageGroup")
+
+            fields_to_check = ["penalize_variation_injection", "penalize_variation_withdrawal", "efficiency_withdrawal"]
+
+            violations = [field for field in fields_to_check if field in user_dict]
+            if violations:
+                raise ValueError(f"In version 8.8, the following values are not allowed: {', '.join(violations)}")
+
+            for field in fields_to_check:
```
MartinBelthle, line 90: I don't see how we can reach this code as you'd raise above

src/antares/craft/service/local_services/models/st_storage.py
```diff
@@ -22,28 +28,107 @@ def _sts_alias_generator(input: str) -> str:
...
+                    _ = STStorageGroup(v)
+                    return v
+                except ValueError:
+                    raise ValueError(f"Group for 8.8 has to be a valid value : {[e.value for e in STStorageGroup]}")
+            if isinstance(v, STStorageGroup):
+                return v.value
+        return str(v) if v is not None else ""
+
+    @field_validator("efficiency_withdrawal", "penalize_variation_injection", "penalize_variation_withdrawal")
+    def check_version_support(cls, v: Union[str, STStorageGroup], info: ValidationInfo) -> Any:
```
MartinBelthle, line 61: We don't use `Union` anymore but this instead `|`
Also, v could be None

src/antares/craft/service/local_services/models/st_storage.py
```diff
@@ -22,28 +28,107 @@ def _sts_alias_generator(input: str) -> str:
...
+        if study_version == STUDY_VERSION_8_8:
+            if v is None or v == "":
+                return STStorageGroup.OTHER1.value
+            if isinstance(v, str):
+                try:
+                    _ = STStorageGroup(v)
+                    return v
+                except ValueError:
+                    raise ValueError(f"Group for 8.8 has to be a valid value : {[e.value for e in STStorageGroup]}")
+            if isinstance(v, STStorageGroup):
```
MartinBelthle, line 56: i really don't like the code in this file but we can tackle this after all the other comments

MartinBelthle (changes requested):
You have a little conflict with `main` branch

tests/antares/services/api_services/test_area_api.py
vargastat: done

src/antares/craft/service/local_services/models/st_storage.py
vargastat: nothing better I found :(

MartinBelthle (approved):
The integration test is also OK :)

:warning don't merge this PR yet as we need the other 9.2 features to be integrated

PR #184, opened 2025-04-29
Release9.2/matrice

berthetclement:
st-storage : 

- Create cluster with default TS values (5 new .txt files)
- Add 5 new "setter" for every TS


src/antares/craft/service/local_services/services/area.py
```diff
@@ -263,6 +264,42 @@ def create_st_storage(
         write_timeseries(self.config.study_path, default_matrix_ones, TimeSeriesFileType.ST_STORAGE_UPPER_RULE_CURVE, area_id, cluster_id=cluster_id)
         # fmt: on
 
+        if self.study_version >= STUDY_VERSION_9_2:
+            write_timeseries(
+                self.config.study_path,
+                default_matrix_zeros,
```
MartinBelthle, line 270: We can simply write an empty matrix for the 5 new matrices (new since v9.2)

tests/antares/services/local_services/test_area.py
```diff
@@ -223,6 +224,76 @@ def test_st_storage_and_ini_have_custom_properties_92(self, local_study_92):
...
+    def test_creation_default_matrices_92(self, tmp_path: Path, local_study_92) -> None:
+        # given
+        st_storage_name = "storage_ts"
+        local_study_92.get_areas()["fr"].create_st_storage(st_storage_name)
+
+        # when
+        storage = local_study_92.get_areas()["fr"].get_st_storages()[st_storage_name]
+
+        # why ?
+        storage.get_pmax_injection()
```
MartinBelthle, line 236: You can simply remove lines from 235 to 248

tests/antares/services/local_services/test_area.py
```diff
@@ -223,6 +224,76 @@ def test_st_storage_and_ini_have_custom_properties_92(self, local_study_92):
...
+        storage.set_cost_withdrawal(matrix)
+        assert storage.get_cost_withdrawal().equals(matrix)
+        storage.set_cost_level(matrix)
+        assert storage.get_cost_level().equals(matrix)
+        storage.set_cost_variation_injection(matrix)
+        assert storage.get_cost_variation_injection().equals(matrix)
+        storage.set_cost_variation_withdrawal(matrix)
+        assert storage.get_cost_variation_withdrawal().equals(matrix)
+
+    def test_update_matrices_wrong_format_92(self, tmp_path: Path, local_study_92) -> None:
```
MartinBelthle, line 279: We should add 2 tests: 
1) Ensures the cost matrices are not written and cannot be `get` with a 8.8 study
2) Ensures the cost matrices cannot be modify with a 8.8 study
In such cases we should raise an explicit error msg

MartinBelthle (changes requested):

src/antares/craft/service/local_services/services/area.py
berthetclement: done with `empty_matrix = pd.DataFrame()`

tests/antares/services/local_services/test_area.py
berthetclement: It's removed

ps : I've found this kind of writing in tests, is it useful?

tests/antares/services/local_services/test_area.py
berthetclement: Done, all new "getter"/"setter" (v9.2) are not allowed by `get_storage_matrix `and `set_storage_matrix `for a v8.8 study

new tests added : 
- `test_creation_matrices_not_allowed_88`
- `test_update_matrices_88_not_allowed`

src/antares/craft/service/local_services/services/st_storage.py
```diff
@@ -87,11 +93,32 @@ def read_st_storages(self) -> dict[str, dict[str, STStorage]]:
 
     @override
     def set_storage_matrix(self, storage: STStorage, ts_name: STStorageMatrixName, matrix: pd.DataFrame) -> None:
+        forbidden_matrices_88 = [
```
MartinBelthle, line 96: Minor but you could put the check inside a function to avoid duplicating code

tests/antares/services/local_services/test_area.py
```diff
@@ -223,6 +224,127 @@ def test_st_storage_and_ini_have_custom_properties_92(self, local_study_92):
 
         assert created_storage.properties == properties
 
+    def test_creation_default_matrices_92(self, tmp_path: Path, local_study_92) -> None:
+        # given
+        st_storage_name = "storage_ts"
+        local_study_92.get_areas()["fr"].create_st_storage(st_storage_name)
+
+        # when
+        storage = local_study_92.get_areas()["fr"].get_st_storages()[st_storage_name]
```
MartinBelthle, line 233: Minor: You could simplify: 
`storage =local_study_92.get_areas()["fr"].create_st_storage(st_storage_name)`
Same for the other tests below

tests/antares/services/local_services/test_area.py
```diff
@@ -223,6 +224,127 @@ def test_st_storage_and_ini_have_custom_properties_92(self, local_study_92):
...
+        # given
+        st_storage_name = "storage_ts"
+        local_study_w_areas.get_areas()["fr"].create_st_storage(st_storage_name)
+
+        # when
+        storage = local_study_w_areas.get_areas()["fr"].get_st_storages()[st_storage_name]
+
+        # then
+        with pytest.raises(
+            ValueError, match="The matrix STStorageMatrixName.COST_INJECTION is not available for study version 8.8"
```
MartinBelthle, line 253: We should log `cost_injection` instead of `STStorageMatrixName.COST_INJECTION`.
You can do that by putting `.value` at the end of the enum when raising the Exception

MartinBelthle (changes requested):
Last comments, pretty minor. After that you can merge

src/antares/craft/service/local_services/services/st_storage.py
vargastat: Done.
I've reassembled the duplicated block of code and I'm doing the check with a `_check_matrix_allowed()` function.

tests/antares/services/local_services/test_area.py
berthetclement: Done, all test simplified

PR #186, opened 2025-05-06
feat(output / api): reworking model output and interaction with the a…

mehdiwahada:
…pi (wip unit and integration tests)

### DIFF
--- a/src/antares/craft/model/output.py
+++ b/src/antares/craft/model/output.py
@@ -18,14 +18,14 @@ import pandas as pd
 from antares.craft.service.base_services import BaseOutputService
 
 
-class MCIndAreas(Enum):
+class MCIndAreasDataType(Enum):
     VALUES = "values"
     DETAILS = "details"
     DETAILS_ST_STORAGE = "details-STstorage"
     DETAILS_RES = "details-res"
 
 
-class MCAllAreas(Enum):
+class MCAllAreasDataType(Enum):
     VALUES = "values"
     DETAILS = "details"
     DETAILS_ST_STORAGE = "details-STstorage"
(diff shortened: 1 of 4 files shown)


tests/antares/services/api_services/test_study_api.py
```diff
@@ -648,6 +657,53 @@ def test_output_get_matrix(self):
...
 
+    def test_output_get_mc_all(self):
+        self.study._areas["area_test"] = self.area
+        self.study._areas["area_test_1"] = self.area_1
+        self.study._outputs["test-output-link"] = self.output_link
+        self.study._outputs["test-output-area"] = self.output_area
+
+        frequency = Frequency(Frequency.ANNUAL.value)
+        with requests_mock.Mocker() as mocker:
+            matrix_link_url = f"https://antares.com/api/v1/studies/{self.study_id}/raw?path=output/{self.output_link.name}/economy/mc-all/links/{self.area.id}/{self.area_1.id}/values-{frequency.value}"
```
sylvlecl, line 668: As briefly discussed, I think we could instead use the endpoint `original-file` in the web implementation, so that we can use the same parsing method in the web implementation as in the local implementation.
(its uses the `read_csv` method of pandas the appropriate parameters in particular to read multiple lines for the column headers)

src/antares/craft/model/output.py
```diff
@@ -128,11 +207,11 @@ def aggregate_areas_mc_ind(
 
         return self._output_service.aggregate_values(self.name, aggregation_entry, "areas", "ind")
 
-    def aggregate_links_mc_ind(
+    def mc_ind_aggregate_links(
         self,
-        query_file: str,
-        frequency: str,
-        mc_years: Optional[list[str]] = None,
+        query_file: MCIndLinksDataType,
```
sylvlecl, line 212: note: to be renamed to data_type too

sylvlecl (commented):
A few comments

sylvlecl (changes requested):
A few more adaptations are needed before merging :)

sylvlecl (approved):

PR #196, opened 2025-06-25
feat(area): implement local deletion

MartinBelthle:


src/antares/craft/model/study.py
```diff
@@ -163,6 +163,13 @@ def delete_area(self, area: Area) -> None:
         # Delete the area
         self._area_service.delete_area(area.id)
         self._areas.pop(area.id)
+        # Delete it from the links
+        links_to_remove = []
+        for link_id, link in self._links.items():
+            if link.area_from_id == area.id or link.area_to_id == area.id:
+                links_to_remove.append(link_id)
+        for link_id in links_to_remove:
+            self._links.pop(link_id)
```
MartinBelthle, line 172: We should also remove them from the fs

PR #279, opened 2025-10-16
Feat/reformate setting terms

mehdiwahada:
How this functionality works : the method replace the actual list of constraint terms by a fresh new one and, **in local**, recreate the constraint dictionary with the id, name, properties and replace the terms in it.

Purpose of the task : This task serves to simplify the binding constraint terms API endpoints by adding one method that could add / update the terms (Don't know about the delete part), thus progressively flushing out individual term manipulation endpoints

Name of the method and how it can be tested : 
- Access : through BindingConstraint method set_terms (-> set_constraint_terms from BindingConstraintService)
- Tests : 
- **(API)** First, create terms and add them to your constraints with set_terms, then, compare the actual constraint terms (get_terms) with the expected list. Second, to check the replace functionality work, do the same with new constraint terms
- **(Local)** Same as API, except you must compare the actual ini_content terms with the expected one

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,3 +1,29 @@
+v0.5.0 (2025-10-13)
+-------------------
+
+### Compatiblity
+* This version fully supports the Simulator v9.3 [251](https://github.com/AntaresSimulatorTeam/antares_craft/pull/251), [277](https://github.com/AntaresSimulatorTeam/antares_craft/pull/277)
+* This version is compatible with AntaresWeb v2.25.0. The previous is not.
+
+### Bug fixes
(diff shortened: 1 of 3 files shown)


tests/integration/test_web_client.py
```diff
@@ -567,9 +569,11 @@ def test_lifecycle(
         study.delete_binding_constraint(constraint_1)
         assert constraint_1.id not in study.get_binding_constraints()
 
-        # tests constraint term deletion
-        constraint_2.delete_term(link_term_2)
-        assert link_term_2.id not in constraint_2.get_terms()
+        # # tests constraint term deletion
```
sylvlecl: indeed we don't need it anymore, we should remove the commented out code

sylvlecl (approved):

PR #398, opened 2026-04-22
fix(local, sts): additional constraints were saved in a wrong format

salemsd:
Fix following a pydantic validation error because of a model mismatch (`STStorageAdditionalConstraintFileData` in AW doesn't contain `name` but it's written in the `.ini` file)

### DIFF
--- a/src/antares/craft/service/local_services/services/st_storage.py
+++ b/src/antares/craft/service/local_services/services/st_storage.py
@@ -211,7 +211,7 @@ class ShortTermStorageLocalService(BaseShortTermStorageService):
             # Round trip for validation
             saved_constraints[constraint.id] = parse_st_storage_constraint_local(content)
 
-            content.pop("name", None) # AW doesn't take name in the INI file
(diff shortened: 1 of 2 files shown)


src/antares/craft/service/local_services/services/st_storage.py
```diff
@@ -208,10 +208,12 @@ def _save_constraints(
         ini_content = {}
         for constraint in constraints.values():
             content = serialize_st_storage_constraint_local(constraint)
-            ini_content[constraint.name] = content
             # Round trip for validation
             saved_constraints[constraint.id] = parse_st_storage_constraint_local(content)
 
+            content.pop("name", None)  # AW doesn't take name in the INI file
```
MartinBelthle: Thanks for the fix :). 

Just for info, this is not really AW that does not take the name in the INI file but rather the Simulator who does not accept it. So you can remove you comment as it's a bit misleading

MartinBelthle (changes requested):

MartinBelthle (approved):

PR #402, opened 2026-05-14
Add local study validation API

mzl2233:
Adds a lightweight `validate_study_local` entry point that checks whether a local study path has the expected directory and readable `study.antares` metadata without loading the full study object. This lets callers quickly skip invalid study paths while keeping the existing `read_study_local` behavior unchanged.

Verification: `python3 -m compileall -q src/antares/craft/service/local_services/factory.py src/antares/craft/model/study.py src/antares/craft/__init__.py`; smoke check with `create_study_local` and `validate_study_local`; `python3 -m pytest tests/antares/services/local_services/test_study_read.py -q`; `ruff check src/antares/craft/service/local_services/factory.py src/antares/craft/model/study.py src/antares/craft/__init__.py`; `git diff --check`.

### DIFF
--- a/src/antares/craft/__init__.py
+++ b/src/antares/craft/__init__.py
@@ -97,6 +97,7 @@ from antares.craft.model.study import (
     import_study_api,
     read_study_api,
     read_study_local,
+    validate_study_local,
 )
 from antares.craft.model.thermal import (
     LawOption,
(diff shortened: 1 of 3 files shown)


MartinBelthle:
The issue is no longer relevant, so does this PR

PR #438, opened 2026-09-22
feat(local constraints terms): checking submitted links and clusters in terms

mehdiwahada:
What does it do ? : Adds a check for the submitted constraints, verifying if the link / thermal cluster datas (for a link, the two areas, for example) exists and returns an exception if not

Why it has been done ? : the constraints are being created without verifying if the area and clusters (or areas if it's a LinkData)

How to test it ? : 
- In the case of a Link, you must create a LinkData variable with two areas not linked together
- In the case of a ThermalCluster, create a term where the area doesn't exist, and then create a term where the area exists, but not the cluster


src/antares/craft/model/study.py
```diff
@@ -323,6 +324,30 @@ def create_multiple_binding_constraints(
         Args:
             data: A dictionary mapping constraint names to tuples of properties and terms.
         """
+
+        # Checking the constraints before creating them, and return an exception if the link / cluster does not exist
+        for constraint_name, (properties, terms) in data.items():
+            for term in terms:
+                if isinstance(term.data, LinkData):
+                    link_id = term.data.area1 + " / " + term.data.area2
+                    if link_id not in self._links:
```
smailio, line 333: _link is a private attribute, you should rather access it through its getter `get_links`. Another solution would be to add a `link_exists(area1, area2)` method, this way the logic to check link existence is centralize. It's up to you.

src/antares/craft/model/study.py
```diff
@@ -323,6 +324,30 @@ def create_multiple_binding_constraints(
         Args:
             data: A dictionary mapping constraint names to tuples of properties and terms.
         """
+
+        # Checking the constraints before creating them, and return an exception if the link / cluster does not exist
+        for constraint_name, (properties, terms) in data.items():
+            for term in terms:
+                if isinstance(term.data, LinkData):
+                    link_id = term.data.area1 + " / " + term.data.area2
```
smailio, line 332: this should be sorted, like it's done in the id property of Link.

src/antares/craft/model/study.py
```diff
@@ -323,6 +324,30 @@ def create_multiple_binding_constraints(
...
+                    if link_id not in self._links:
+                        raise BindingConstraintCreationError(
+                            constraint_name=constraint_name, message=f"Link '{link_id}' does not exist"
+                        )
+
+                if isinstance(term.data, ClusterData):
+                    area_id = term.data.area
+                    cluster_id = term.data.cluster
+                    if area_id not in self._areas:
```
smailio: _areas is also private

src/antares/craft/model/study.py
```diff
@@ -304,6 +305,27 @@ def create_binding_constraint(
         Returns:
             The newly created binding constraint.
         """
+        if terms is not None:
+            for term in terms:
+                if term.data is not None and isinstance(term.data, LinkData):
```
smailio: term.data can't be None, can it ?

src/antares/craft/model/study.py
```diff
@@ -304,6 +305,27 @@ def create_binding_constraint(
...
+                            constraint_name=name, message=f"Link '{link_id}' does not exist"
+                        )
+                if term.data is not None and isinstance(term.data, ClusterData):
+                    area_id = term.data.area
+                    cluster_id = term.data.cluster
+                    if not self.area_exists(area_id):
+                        raise BindingConstraintCreationError(
+                            constraint_name=name, message=f"Area '{area_id}' does not exist"
+                        )
+                    else:
```
smailio: no need fot the else block as if you enter the if aboce it raises

src/antares/craft/model/study.py
```diff
@@ -304,6 +305,27 @@ def create_binding_constraint(
         Returns:
             The newly created binding constraint.
         """
+        if terms is not None:
+            for term in terms:
+                if term.data is not None and isinstance(term.data, LinkData):
+                    if not self.link_exists(term.data.area1, term.data.area2):
+                        link_id = " / ".join(sorted((term.data.area1, term.data.area2)))
```
smailio: no need for this local variable, just write bellow

`message=f"Link '{term.data.area1}' - '{term.data.area1}' does not exist"

src/antares/craft/model/study.py
```diff
@@ -323,6 +347,30 @@ def create_multiple_binding_constraints(
...
+                    area_id = term.data.area
+                    cluster_id = term.data.cluster
+                    if not self.area_exists(area_id):
+                        raise BindingConstraintCreationError(
+                            constraint_name=constraint_name, message=f"Area '{area_id}' does not exist"
+                        )
+                    else:
+                        if not self.cluster_exists(area_id, cluster_id):
+                            raise BindingConstraintCreationError(
+                                constraint_name=constraint_name, message=f"Cluster '{cluster_id}' does not exist"
```
smailio: this logic is the same than above right ? You can extract this logic in a validate_constraint_term function (if you can think of a  better name use it)

src/antares/craft/model/study.py
mehdiwahada: That's true, indeed. My bad.

src/antares/craft/model/study.py
mehdiwahada: Good idea. Should I use the area_exists into link_exists btw ?

smailio (approved):

PR #439, opened 2026-09-23
feat(api constraint_terms): added api testing on constraint terms checking

mehdiwahada:


tests/integration/test_web_client.py
```diff
@@ -469,6 +469,45 @@ def test_lifecycle(
...
+            study.create_multiple_binding_constraints(
+                {"bc_4": (BindingConstraintProperties(), [ConstraintTerm(data=invalid_area_id_cluster_data, weight=1)])}
+            )
+
+        # non existing cluster
+        with pytest.raises(
+            BindingConstraintCreationError,
+            match="Could not create the binding constraint 'bc_4'",
+        ):
+            study.create_multiple_binding_constraints(
```
MartinBelthle: We should also test the `create_binding_constraint` method. You can replace this last test with a test of the other method

MartinBelthle (changes requested):

tests/integration/test_web_client.py
mehdiwahada: I admit I don't understand. When it comes to non existent areas or clusters, I should replace the create_multiple_... by create_binding_constraint ?

tests/integration/test_web_client.py
```diff
@@ -469,6 +469,45 @@ def test_lifecycle(
...
+                {"bc_4": (BindingConstraintProperties(), [ConstraintTerm(data=invalid_area_id_cluster_data, weight=1)])}
+            )
+
+        # non existing cluster
+        with pytest.raises(
+            BindingConstraintCreationError,
+            match="Could not create the binding constraint 'bc_4'",
+        ):
+            study.create_binding_constraint(
+                name="bc_4", less_term_matrix=pd.DataFrame(data=np.zeros((8760, 1))), properties=BindingConstraintProperties(), terms=[ConstraintTerm(data=invalid_cluster_data, weight=1)]
```
MartinBelthle: You can remove `less_term_matrix` and `properties` they are optional and not related to the test

MartinBelthle (changes requested):
Seems you should reformat your code. After this we'll merge the PR

